PR #53, opened 2021-12-22
Replace hashed input UtxoId with pair of (txID,outputIndex)

rakita:
Change related to spec: https://github.com/FuelLabs/fuel-specs/pull/253

Additionally i run clippy and fixed its warning. And there was one error that clippy gave:
```
error: written amount is not handled. Use `Write::write_all` instead
   --> src/receipt.rs:866:9
    |
866 |         instance.write(bytes)?;
    |         ^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::unused_io_amount)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount
```

`is_empty` and `unwrap_or_else are clippy requests (she is very needy).

When this is merged i will go through other projects and fix if there is something broken.

### DIFF
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -871,7 +871,7 @@ impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
 
-        instance.write(bytes)?;
+        let _ = instance.write(bytes)?;
 
         Ok(instance)
     }
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -871,7 +871,7 @@ impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
 
-        instance.write_all(bytes)?;
+        instance.write(bytes)?;
 
         Ok(instance)
     }
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -871,6 +871,8 @@ impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
 
+        // We are sure that all needed bytes are written or error would happen.
+        // unused let is here to silence clippy warning for this check.
         let _ = instance.write(bytes)?;
 
         Ok(instance)
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -369,10 +369,10 @@ impl Receipt {
     }
 
     pub const fn is_empty(&self) -> Option<bool> {
-        if let Some(len) = self.len() {
-            Some(len == 0)
-        } else {
-            None
+        match self.len() {
+            Some(0) => Some(true),
+            Some(_) => Some(false),
+            None => None,
         }
     }
 
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -368,6 +368,14 @@ impl Receipt {
         }
     }
 
+    pub const fn is_empty(&self) -> Option<bool> {
+        if let Some(len) = self.len() {
+            Some(len == 0)
+        } else {
+            None
+        }
+    }
+
     pub const fn digest(&self) -> Option<&Bytes32> {
         match self {
             Self::ReturnData { digest, .. } => Some(digest),
@@ -863,7 +871,7 @@ impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
 
-        instance.write(bytes)?;
+        instance.write_all(bytes)?;
 
         Ok(instance)
     }
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -4,7 +4,6 @@ use itertools::Itertools;
 
 use std::convert::TryFrom;
 use std::io::Write;
-use std::sync::Arc;
 use std::{io, mem};
 
 mod id;
@@ -99,8 +98,6 @@ pub enum Transaction {
         metadata: Option<Metadata>,
     },
 }
-
-pub type ArcTx = Arc<Transaction>;
 
 impl Default for Transaction {
     fn default() -> Self {
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -4,6 +4,7 @@ use itertools::Itertools;
 
 use std::convert::TryFrom;
 use std::io::Write;
+use std::sync::Arc;
 use std::{io, mem};
 
 mod id;
@@ -14,7 +15,7 @@ mod types;
 mod validation;
 
 pub use metadata::Metadata;
-pub use types::{Input, Output, Witness};
+pub use types::{Input, Output, UtxoId, Witness};
 pub use validation::ValidationError;
 
 const WORD_SIZE: usize = mem::size_of::<Word>();
@@ -41,6 +42,9 @@ const TRANSACTION_CREATE_FIXED_SIZE: usize = WORD_SIZE // Identifier
     + WORD_SIZE // Outputs size
     + WORD_SIZE // Witnesses size
     + Salt::LEN; // Salt
+
+/// Identification of transaction (also called transaction hash)
+pub type TxId = Bytes32;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 enum TransactionRepr {
@@ -95,6 +99,8 @@ pub enum Transaction {
         metadata: Option<Metadata>,
     },
 }
+
+pub type ArcTx = Arc<Transaction>;
 
 impl Default for Transaction {
     fn default() -> Self {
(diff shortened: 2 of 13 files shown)


Voxelot:
I'd like to see a feature branch on fuel-vm and fuel-core with these changes working before merging. That way we don't wind up incrementally publishing fuel-tx as we troubleshoot.

src/receipt.rs
```diff
@@ -368,6 +368,14 @@ impl Receipt {
         }
     }
 
+    pub const fn is_empty(&self) -> Option<bool> {
+        if let Some(len) = self.len() {
+            Some(len == 0)
+        } else {
+            None
+        }
```
vlopes11, line 376: I think a one-liner here would be more clear on what we are doing

```suggestion
        self.len().map(|l| l == 0)
```

src/receipt.rs
```diff
@@ -863,7 +871,7 @@ impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
 
-        instance.write(bytes)?;
+        instance.write_all(bytes)?;
```
vlopes11, line 874: We shouldn't [write_all](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all) because the bytes buffer might be bigger than the required.

I honestly think this clipper lint is misleading. We might as well use variable size buffer and eventually use the remainder buffer for other data. If we do so, we will have [ErrorKind::Interrupted](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Interrupted) when using `write_all` while `write` will just get the bytes required by this instance.

I personally don't like clipper a lot because of many small nits like this one.

src/transaction.rs
```diff
@@ -42,6 +42,9 @@ const TRANSACTION_CREATE_FIXED_SIZE: usize = WORD_SIZE // Identifier
     + WORD_SIZE // Witnesses size
     + Salt::LEN; // Salt
 
+/// Identification of transaction (also called transaction hash)
+pub type TxId = Bytes32;
```
vlopes11, line 46: This should be moved to `fuel-types` so other crates can have access to it.

src/transaction/id.rs
```diff
@@ -51,7 +51,7 @@ impl Transaction {
                 ..
             } = input
             {
-                utxo_id.iter_mut().for_each(|b| *b = 0);
+                *utxo_id = UtxoId::default();
```
vlopes11, line 54: This will lead to performance degradation. The previous code will just `memset` while the new code will potentially allocate+copy. Probably a clipper nit?

src/transaction/types/input.rs
```diff
@@ -260,7 +263,10 @@ impl io::Write for Input {
                 let (size, predicate_data, _) = bytes::restore_raw_bytes(buf, predicate_data_len)?;
                 n += size;
 
-                let utxo_id = utxo_id.into();
+                let utxo_id = UtxoId {
```
vlopes11, line 266: We need constructors to encapsulate the behavior of `UtxoId` instead of coupling the code here.

src/transaction/types/utxo_id.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// Identification of unspend transaction output.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(
+    feature = "serde-types-minimal",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+pub struct UtxoId {
+    /// transaction id
+    pub tx_id: TxId,
```
vlopes11, line 15: The crate don't use `pub` accessors because this leads to coupled code. Use `pub const fn x(&self) -> X` instead

src/transaction/types/utxo_id.rs
```diff
@@ -0,0 +1,113 @@
...
+                    .map(hex_val)
+                    .flatten()
+                    .ok_or(ERR)?,
+            }
+        };
+        Ok(utxo_id)
+    }
+}
+
+const fn hex_val(c: u8) -> Option<u8> {
```
vlopes11, line 68: Duplicated code to `fuel-types`. Should not be here after `TxId` is moved there

tests/bytes.rs
```diff
@@ -82,7 +82,10 @@ fn input() {
 
     assert_encoding_correct(&[
         Input::coin(
-            rng.gen(),
+            UtxoId {
```
vlopes11, line 85: `UtxoId` should implement [Distribution](https://docs.rs/rand/0.8.4/rand/distributions/trait.Distribution.html) when `random` feature is active. This way you can keep the `rng.gen()`

vlopes11 (changes requested):
Couple of nits/discussions

src/transaction/id.rs
rakita: UtxoId now has two field so i used default. But you got me interested in optimization side and for [u8;32] it is not even memset but moveups, and with optimizer turned on it does not matter, it is same for assign, iter and default: https://rust.godbolt.org/z/xhjaqjxWe

src/receipt.rs
rakita: Tried it before but `is_empty` is const function so i can't call non const `map`.
maybe something like is better:
```rust
        match self.len() {
            Some(0) => Some(true),
            Some(_) => Some(false),
            None => None,
        }
```
wdyt?

src/receipt.rs
rakita: I see, will be reverted.

src/transaction.rs
rakita: `TxId` is more transaction thing, but sure I will move it. On another topic, wdyt about moving `UtxoId` to `fuel-types` it has same feel as `TxId`

tests/bytes.rs
rakita: This is nice way, i did implemented it but it is not hidden behind `random` feature, i tried to correct it but it took more time then needed and just left it, it seems `Witness` has same thing.

src/transaction.rs
Voxelot: It's not totally clear to me where the dividing line is between tx related types and fuel-types. Would it be confusing to us down the road if we start moving bits of pieces of tx's into fuel-types?

rakita: You can probably argue that `TxId` and `UtxoId` are native types and that they belong to `fuel-types`. They can be reexported in `fuel-tx`.
For me personally, i dont really care how I am going to include it, i care about merging this and continuing to next thing :)

adlerjohn: Seconding what @Voxelot. IMO, it wouldn't hurt to first keep these new types here, _then if need arises_ move them to `fuel-types`.

adlerjohn: One argument to move it immediately is that there's code duplication already: https://github.com/FuelLabs/fuel-tx/pull/53#discussion_r778720939. Avoiding code duplication is probably a `Good Thing`, but can that also be accomplished by exporting the duplicate function? I guess that would unclean, as it's a library of types and not functions.

src/transaction/types/utxo_id.rs
rakita: Removed duplicated part. Found that there is `u8::from_str_radix` that does the job.

src/transaction.rs
rakita: Removed duplicated part, it is now using std function `u8::from_str_radix`. And had a litlle bug there, that is fixed in last commit.

src/receipt.rs
```diff
@@ -863,7 +871,7 @@ impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
 
-        instance.write(bytes)?;
+        let _ = instance.write(bytes)?;
```
adlerjohn, line 874: Probably left from debugging?

```suggestion
        instance.write(bytes)?;
```

rakita: Necesarry sacrifice to god Clippy.
If ommited it will report `error: written amount is not handled.`

adlerjohn: Is that indicative that it should be handled then?? 😂 What does `write` return? A `Result`?

rakita: God Victor accused god Clippy of being inacurate and clippy was silenced. Victor blessings can be read here: https://github.com/FuelLabs/fuel-tx/pull/53#discussion_r778759630
it returns number of written bytes, in this particular case we are sure that all bytes are going to be written or error will be returned. Error handling part is here `)?;`

adlerjohn: Ah got it 😂. Can you document this in a comment? Otherwise in the future someone who's not aware of the context might think to delete the `let _ = ` and/or handle the error redundantly.

rakita: sure, comment is added.

adlerjohn (approved):
LGTM, deferring to one of @Voxelot or @vlopes11 for 🦀 review

Voxelot (approved):

rakita:
@adlerjohn @Voxelot i will need you to approve this once again, nothing has changed.

Voxelot (approved):

PR #57, opened 2021-12-27
Add `secp256k1` signature for `Input::Coin`

vlopes11:
Add recoverable signature to input coins.

The signed message is the transaction hash. Since it can be signed only
after the transaction is built, the coin input can only be appended
before it is signed since it will define the transaction hash.

This created the need to include a `TransactionBuilder` that will accept
the arguments of the transaction, and sign the coin inputs after the
transaction id is ready.

Also, the signed transaction cannot contain the witnesses.

These changes are breaking regarding the tests; they now should use the
builder structure when input signatures are to be validated.


rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+max_width = 120
+wrap_comments = true
```
Dentosal, line 2: This PR contains quite a lot of formatting-only changes. This complicates the review unnecessarily, and usually I would like to review these separately.

src/transaction/validation.rs
```diff
@@ -1,42 +1,55 @@
...
     ) -> Result<(), ValidationError> {
         match self {
             Self::Coin { predicate, .. } if predicate.len() > MAX_PREDICATE_LENGTH as usize => {
                 Err(ValidationError::InputCoinPredicateLength { index })
             }
 
-            Self::Coin { predicate_data, .. }
-                if predicate_data.len() > MAX_PREDICATE_DATA_LENGTH as usize =>
-            {
+            Self::Coin { predicate_data, .. } if predicate_data.len() > MAX_PREDICATE_DATA_LENGTH as usize => {
```
Dentosal, line 31: These `as usize` casts of `u64` feel weird to me, as if the constant is `> usize::MAX` it will silently truncate. If the constant is length of something, and often cast to `usize`, changing the type of the constant would seem cleaner.

Dentosal (commented):
Generally LGTM, but not too confident on this so deferring to other reviewers

rustfmt.toml
vlopes11: Agreed, removed

src/transaction/validation.rs
vlopes11: This is not introduced in this PR. However, the constants must be consistent with the runtime requirements.

src/transaction.rs
```diff
@@ -303,12 +304,84 @@ impl Transaction {
...
+            owner,
+            amount,
+            color,
+            witness_index,
+            maturity,
+            predicate,
+            predicate_data,
+        );
+
+        self._add_witness(Witness::default());
```
Voxelot, line 346: This seems inefficient because all of these witnesses would be duplicates if all coins have the same owner field. One major advantage of the witness index is that it allows inputs to share common witnesses, instead of requiring a witness on the coin input type.

src/transaction/id.rs
```diff
@@ -58,39 +73,42 @@ impl Transaction {
...
+                *amount = 0;
+                color.iter_mut().for_each(|b| *b = 0);
+            }
 
-                _ => (),
-            });
+            _ => (),
+        });
+
+        witnesses.clear();
```
Voxelot, line 99: This is already done on line 45: `self.set_witnesses(vec![]);`

src/transaction/internals.rs
```diff
@@ -0,0 +1,54 @@
+use crate::{Input, Output, Transaction, Witness};
+
+#[cfg(feature = "internals")]
```
Voxelot, line 3: Does this really need to be under a feature flag? I feel like we should be more opinionated about what are standard public apis are for to reduce feature flag complexity.

src/transaction/types/witness.rs
```diff
@@ -73,12 +76,20 @@ impl Distribution<Witness> for Standard {
...
+impl bytes::SizedBytes for Witness {
+    fn serialized_size(&self) -> usize {
+        WORD_SIZE + bytes::padded_len(self.data.as_slice())
+    }
+}
+
+#[cfg(feature = "std")]
 impl io::Read for Witness {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         bytes::store_bytes(buf, self.data.as_slice()).map(|(n, _)| n)
```
Voxelot, line 88: While our standard interfaces for serialization use std::io, none of this logic seems like it should require `std`. It's just converting a slice of bytes into an array.

Voxelot: Not saying that should be done in this PR, but maybe a good task for people looking for open source contribution work.

src/transaction/types/input.rs
```diff
@@ -174,8 +194,32 @@ impl Input {
             _ => None,
         }
     }
+
+    #[cfg(feature = "std")]
+    pub fn validate_signature(txhash: &Bytes32, owner: &Address, witness: &Witness) -> bool {
```
Voxelot, line 199: Why not use `&self.owner` instead?

src/transaction/validation.rs
```diff
@@ -65,7 +88,22 @@ impl Output {
 }
 
 impl Transaction {
+    #[cfg(feature = "std")]
     pub fn validate(&self, block_height: Word) -> Result<(), ValidationError> {
+        self._validate(block_height, true)
+    }
+
+    #[cfg(feature = "std")]
+    pub fn validate_without_signature(&self, block_height: Word) -> Result<(), ValidationError> {
```
Voxelot, line 97: Is there a use-case where we'd want to validate just the witnesses and nothing else?

src/transaction/validation.rs
```diff
@@ -65,7 +88,22 @@ impl Output {
 }
 
 impl Transaction {
+    #[cfg(feature = "std")]
     pub fn validate(&self, block_height: Word) -> Result<(), ValidationError> {
+        self._validate(block_height, true)
+    }
+
+    #[cfg(feature = "std")]
```
Voxelot, line 96: Seems like we should be able to validate without `std` if we aren't checking signatures?

src/transaction/validation.rs
adlerjohn: Yes: signatures can be validated in parallel, on a GPU even (see: Solana). In such a case, signature verification should be independent from everything else. The owner of each input is provided in the tx, so you can proceed optimistically even without verifying signatures.

src/transaction.rs
vlopes11: That is true. This will involve some state tracking in the transaction builder to construct the witnesses in an optimal way.

The current implementation won't affect the correctness of the protocol. We can create this improvement as a separate PR/issue: https://github.com/FuelLabs/fuel-tx/issues/72

src/transaction/id.rs
vlopes11: True, added this commit to not waste the allocated capacity of the vector:

https://github.com/FuelLabs/fuel-tx/pull/57/commits/b49ea0221c8f3d681dad286e2c8d71eb31f9190f

src/transaction/internals.rs
vlopes11: This `internals` flag is used for methods that shouldn't be exposed by the API. I think we should avoid creating many functions that shouldn't be called by the user. The inputs/outputs/witnesses are coupled and should be manipulated via the API of the builder and of the transaction itself

src/transaction/types/witness.rs
vlopes11: Blanket implementations would probably bring the need to copy on serialization. It will involve a complicated scheme to have a blanket zero-copy `io` implementation

src/transaction/types/input.rs
vlopes11: We can't always have `self.owner` because `Input` also has the variant `Contract`. If we take `self`, the user could call `validate_signature` also for a contract, and that wouldn't validate anything. The intent of taking the `owner` instead of `self` is to translate this restriction clearly in the API

src/transaction/validation.rs
vlopes11: Added API to validate with or without signature:

https://github.com/FuelLabs/fuel-tx/pull/57/commits/a77f4e84eb40e75a806e03c41ce44285b6a18bdc

src/transaction/validation.rs
vlopes11: Fixed in https://github.com/FuelLabs/fuel-tx/pull/57/commits/a77f4e84eb40e75a806e03c41ce44285b6a18bdc

src/transaction/types/input.rs
vlopes11: API changed in https://github.com/FuelLabs/fuel-tx/commit/a77f4e84eb40e75a806e03c41ce44285b6a18bdc

src/transaction.rs
Voxelot: Also I think the sdk won't be using this, so it's not urgent.

src/transaction/internals.rs
Voxelot: It looks like this feature flag isn't used at all anywhere, should we just remove this?

src/transaction/types/witness.rs
Voxelot: Looking at `fuel-types::bytes` it seems like they could be easily made no-std compatible if we just used a different error type. Resolving this for now since it seems like something we could fix in fuel-types.

src/transaction/internals.rs
vlopes11: I think @digorithm was using that for some test purposes

vlopes11: In that case, the only usage is for integrated tests purposes: https://github.com/FuelLabs/fuel-tx/blob/a77f4e84eb40e75a806e03c41ce44285b6a18bdc/test-helpers/Cargo.toml#L9

src/transaction/validation.rs
```diff
@@ -86,49 +171,64 @@ impl Transaction {
...
                 .filter_map(|output| match output {
-                    Output::Variable { color, .. } if input_color == &color => Some(()),
+                    Output::Variable { color, .. } if input_color == color => Some(()),
                     _ => None,
                 })
                 .count()
                 > 1
             {
-                Err(ValidationError::TransactionOutputVariableColorDuplicated)?
+                return Err(ValidationError::TransactionOutputVariableColorDuplicated);
```
Voxelot, line 205: Unrelated to this PR, but I think this validation rule may be incorrect. 

If there are multiple TRO calls of the same asset id, you'd need to have duplicate colors since each TRO can only set a variable output once.  Also, users with variable outputs shouldn't need to set to color in advance anyways since it will be zeroed out prior to execution.

cc: @adlerjohn

test-helpers/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "fuel-tx-test-helpers"
+version = "0.1.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+
```
Voxelot, line 6: Unless we add a proper license and version then we should exclude this from being published.

```suggestion
publish = false
```

test-helpers/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+[dependencies]
+fuel-crypto = { version = "0.2", default-features = false, features = ["random"] }
+fuel-tx = { path = "../../fuel-tx", default-features = false, features = ["alloc","builder","internals","random"] }
```
Voxelot, line 9: path based normal deps aren't compatible with cargo publish.

tests/valid_cases/input.rs
```diff
@@ -1,103 +1,187 @@
...
+        utxo_id,
+        amount,
+        color,
+        maturity,
         vec![0u8; MAX_PREDICATE_LENGTH as usize],
         vec![],
     )
-    .validate(1, &[], witnesses.as_slice())
-    .unwrap();
+    .expect("Failed to validate transaction");
```
Voxelot, line 83: Each of these should probably be broken up into separate unit tests, or at least label what is being tested in each case. There should be some kind of explicit description of each scenario under test and what the intended outcome is beyond just test code itself.

src/transaction.rs
```diff
@@ -307,13 +315,88 @@ impl Transaction {
...
+        inputs
+            .iter()
+            .filter_map(|input| match input {
+                Input::Coin {
+                    owner,
+                    witness_index,
+                    ..
+                } if owner == &pk => Some(*witness_index as usize),
+                _ => None,
+            })
```
Voxelot, line 392:
```suggestion
            })
            .dedupe()
```

src/transaction.rs
```diff
@@ -176,16 +178,30 @@ impl Transaction {
     }
 
     pub fn input_colors(&self) -> impl Iterator<Item = &Color> {
```
Voxelot, line 180: Why break this API and make a new one to do the same thing?

src/transaction/id.rs
```diff
@@ -362,18 +372,32 @@ mod tests {
                                     } => invert(static_contracts.first_mut().unwrap()),
```
Voxelot, line 372: This is duplicated on line 381

src/transaction/id.rs
```diff
@@ -362,18 +372,32 @@ mod tests {
...
+                                if !static_contracts.is_empty() {
+                                    assert_id_ne(&tx, |t| match t {
+                                        Transaction::Create {
+                                            static_contracts, ..
+                                        } => invert(static_contracts.first_mut().unwrap()),
+                                        _ => unreachable!(),
+                                    });
+                                }
+
+                                if !storage_slots.is_empty() {
```
Voxelot, line 390: This shouldn't be wrapped inside of `if !static_contracts.is_empty() {` on line 368, the first iteration of the outer loop will always have static_contracts as empty.

tests/valid_cases/transaction.rs
```diff
@@ -111,755 +148,742 @@ fn maturity() {
...
+    while builder.witnesses().len() < MAX_WITNESSES as usize {
+        builder.add_witness(generate_bytes(rng).into());
+    }
+
+    builder
+        .finalize()
+        .validate(block_height)
+        .expect("Failed to validate transaction");
+
+    let mut builder = TransactionBuilder::create(MAX_WITNESSES - 1, rng.gen(), vec![], vec![]);
```
Voxelot, line 194: For readability we should have these as separate tests or at least a comment separator between tests explaining what is about to be tested.

Voxelot:
Think I'm pretty much ready to approve once a few of my more major comments are addressed (I think there must've been some kind of merge glitch on the transaction::id tests)

src/transaction/validation.rs
vlopes11: https://github.com/FuelLabs/fuel-tx/issues/82

test-helpers/Cargo.toml
vlopes11: This isn't intended to be published

tests/valid_cases/input.rs
vlopes11: https://github.com/FuelLabs/fuel-tx/issues/83

src/transaction.rs
vlopes11: This wasn't changed: https://github.com/FuelLabs/fuel-tx/blob/d37cb3b0cdc02ed120f7ccbe5648fb8c8205a228/src/transaction.rs#L178

src/transaction/validation.rs
```diff
@@ -29,6 +90,8 @@ impl Input {
                 Err(ValidationError::InputCoinWitnessIndexBounds { index })
             }
 
+            Self::Coin { .. } => Ok(()),
```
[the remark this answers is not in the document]
vlopes11: Its separated because of the `TODO` there: https://github.com/FuelLabs/fuel-tx/blob/d37cb3b0cdc02ed120f7ccbe5648fb8c8205a228/src/transaction/validation.rs#L47-L49

src/transaction.rs
vlopes11: *fixed in https://github.com/FuelLabs/fuel-tx/pull/57/commits/bc11f3c307175412c752b7ede457213521e77ae0

tests/valid_cases/transaction.rs
vlopes11: Added comments in: https://github.com/FuelLabs/fuel-tx/pull/57/commits/b0f2e785627fc60cedd531ddd985a7a79e835ba9

Also related: https://github.com/FuelLabs/fuel-tx/issues/83

src/transaction.rs
Voxelot: This API used to dedupe the colors, but now the dedupe logic was moved to a separate api.

test-helpers/Cargo.toml
Voxelot: yeah I could see this builder being useful downstream for VM testing, it's very similar to the `TestBuilder` I added there while working on the `TR` and `TRO` opcodes.

src/transaction/internals.rs
Voxelot: maybe we could consolidate these flags under `test-helpers` or something instead of making a bunch of different flags for things that are just dev-deps?

Voxelot (approved):

Voxelot:
left a couple nits but not blockers in any way. Approoooved!

PR #70, opened 2022-01-17
Change witness size validation

digorithm:


adlerjohn (approved):

tests/valid_cases/transaction.rs
```diff
@@ -727,7 +727,7 @@ fn create() {
             vec![],
         )],
         vec![Output::change(rng.gen(), rng.next_u64(), Color::default())],
-        vec![vec![0xfau8; 1 + CONTRACT_MAX_SIZE as usize / 4].into()],
+        vec![vec![0xfau8; 1 + CONTRACT_MAX_SIZE as usize].into()],
```
Voxelot, line 730: Seems like the original intention behind this test was for the array to be max_length + 1. Since `0xfau8` is 2 bytes, we could `/ 2` here but not a big deal either way.

Voxelot (approved):

tests/valid_cases/transaction.rs
adlerjohn: Copying from Slack:

> Oh wait
> `0xfau8` isn't 2 bytes, that's why
> It's a `0xfa` `u8`!

PR #76, opened 2022-02-03
v0.4.0 Release

Voxelot:
Release storage slot initialization

also improve speed of tag validation


adlerjohn (approved):
I'll trust that you tested the toml check. utACK.

Voxelot:
It was tested on the dummy repo, so it would only be an issue if I copied something over wrong.

PR #94, opened 2022-03-26
add new error cases and validation rules

Voxelot:
Adds several new error types as well as a validation rule for coin outputs.

This is largely to address the tech debt identified here: https://github.com/FuelLabs/fuel-vm/pull/93#pullrequestreview-921424472

### DIFF
--- a/src/transaction/validation/error.rs
+++ b/src/transaction/validation/error.rs
@@ -1,5 +1,6 @@
 use core::fmt;
 
+use fuel_types::AssetId;
 #[cfg(feature = "std")]
 use std::{error, io};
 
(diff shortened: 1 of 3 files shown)


src/transaction/validation/error.rs
```diff
@@ -28,16 +51,40 @@ pub enum ValidationError {
...
+    InsufficientInputAmount {
+        /// The asset id being spent
+        asset: AssetId,
+        /// The amount expected by a coin output
+        expected: u64,
+        /// The total amount provided by coin inputs
+        provided: u64,
+    },
+    /// The user provided amounts for coins or gas prices that caused an arithmetic
+    /// overflow.
```
adlerjohn, line 86: Ahhh this should actually be in the specs!

adlerjohn (approved):

Issue #88, opened 2022-03-03
InstructionResult Inconsistencies


Voxelot:
InstructionResult is used inconsistently between panic and script result types:

```
    Panic {
        id: ContractId,
        reason: Word,
        pc: Word,
        is: Word,
    },
...
    ScriptResult {
        result: InstructionResult,
        gas_used: Word,
    },
```

The reason for the panic receipt is always an encoded `InstructionResult`, so it would be more user-friendly if the panic receipt used the InstructionResult rather than the encoded form.

adlerjohn:
Hmmmm. Could you clarify? The specs for the two receipts are here: https://github.com/FuelLabs/fuel-specs/blob/38d259f1a239908c01f1471d8efe8c9e5a58e80f/specs/vm/opcodes.md#reading-guide

The ABI specs are here: https://github.com/FuelLabs/fuel-specs/blob/38d259f1a239908c01f1471d8efe8c9e5a58e80f/specs/protocol/abi.md#panic-receipt and here: https://github.com/FuelLabs/fuel-specs/blob/38d259f1a239908c01f1471d8efe8c9e5a58e80f/specs/protocol/abi.md#scriptresult-receipt.

It appears the ABI specs are inconsistent with the VM specs!

Voxelot:
I see, so the panic receipt is only meant to include the `PanicReason` and not also include the encoded instruction. That makes sense since the ScriptResult will already include the instruction result. In that case it seems like we should have this:

```
    Panic {
        id: ContractId,
        reason: PanicReason,
        pc: Word,
        is: Word,
    },
...
    ScriptResult {
        result: InstructionResult,
        gas_used: Word,
    },
```

The VM is incorrectly setting up the panic receipt here (which is encoding the actual instruction along with panic reason): https://github.com/FuelLabs/fuel-vm/blob/master/src/interpreter/flow.rs#L210

Voxelot:
After reviewing the specs and discussing more, ScriptResult is meant to be a boolean result, not an encoded panic reason / instruction. We should update the `ScriptResult` result to be a boolean or word and set it based on the updated specs here: https://github.com/FuelLabs/fuel-specs/issues/289

PR #99, opened 2022-03-31
Receipt Error Improvements

Voxelot:
This PR makes necessary changes in order for the fuel_tx::Receipt type able to reflect the recent spec change of ScriptResult reporting Reverts as a transaction failure.

Replaced usage of InstructionResult for ScriptResult `result` field because:
- InstructionResult includes a panic reason, and it's counter-intuitive to use a type called PanicReason for representing a success. This led to obscure logs for users where they'd see a `PanicReason::RESERV00` which could either mean success or revert.
- While there is a PanicReason::Revert, this would also be misleading since a revert can never cause a panic (unless it was out of gas, but then the panic reason should be OutOfGas, not revert).
- InstructionResult is also incompatible with fully representing a revert. This is because `rA` for a revert requires a full word, so it's not possible to pack both variants of InstructionResult (from panic) and `rA` (from revert) into a single word field for result.

ScriptResult.result() now falls in line closer with the specs:
  - 0 == Success
  - 1 == Revert
  - 2 == Panic
  - anything else == Generic Error

This is the most transparency ScriptResult is able to provide using only a single u64 without potential loss of data. A new ScriptExecutionResult type is used to alias these result values to increase transparency for downstream Rust consumers.

The Panic receipt was also modified to use `InstructionResult` instead of `Word` for its' `reason` field. The VM was already encoding an InstructionResult into Panic receipts, but because the reason field used a `Word`, the InstructionResult and panic reason was opaque. As an example, when debugging panic receipts a user would see something like `Panic { reason: 18446744073709, .. }`, which was not helpful and didn't leverage existing available types like InstructionResult or PanicReason. This resolves any confusion that may arise from how to parse the reason field on a Panic, since the reason is no longer specified (https://github.com/FuelLabs/fuel-specs/pull/256) and a Word provided no guarantees whether the VM is using an `InstructionResult`, `PanicReason`, or some other kind of arbitrary value.

fixes: #88

### DIFF
--- a/src/receipt/receipt_std.rs
+++ b/src/receipt/receipt_std.rs
@@ -347,7 +347,7 @@ impl io::Write for Receipt {
 
                 let result = ScriptExecutionResult::from(result);
 
-                *self = Self::script_result(result.into(), gas_used);
+                *self = Self::script_result(result, gas_used);
             }
         }
 
--- a/src/receipt/receipt_std.rs
+++ b/src/receipt/receipt_std.rs
@@ -170,7 +170,7 @@ impl io::Read for Receipt {
             Self::ScriptResult { result, gas_used } => {
                 let buf = bytes::store_number_unchecked(buf, ReceiptRepr::ScriptResult as Word);
 
-                let result: Word = (*result).into();
+                let result = Word::from(*result);
                 let buf = bytes::store_number_unchecked(buf, result);
 
                 bytes::store_number_unchecked(buf, *gas_used);
--- a/src/receipt/receipt_std.rs
+++ b/src/receipt/receipt_std.rs
@@ -4,6 +4,7 @@ use fuel_asm::InstructionResult;
 use fuel_types::bytes::{self, SizedBytes, WORD_SIZE};
 use fuel_types::Word;
 
+use crate::receipt::script_result::ScriptExecutionResult;
 use std::io::{self, Write};
 
 impl io::Read for Receipt {
@@ -169,7 +170,7 @@ impl io::Read for Receipt {
             Self::ScriptResult { result, gas_used } => {
                 let buf = bytes::store_number_unchecked(buf, ReceiptRepr::ScriptResult as Word);
 
-                let result = Word::from(*result);
+                let result: Word = (*result).into();
                 let buf = bytes::store_number_unchecked(buf, result);
 
                 bytes::store_number_unchecked(buf, *gas_used);
@@ -258,7 +259,7 @@ impl io::Write for Receipt {
 
                 let id = id.into();
 
-                *self = Self::panic(id, reason, pc, is);
+                *self = Self::panic(id, InstructionResult::from(reason), pc, is);
             }
 
             ReceiptRepr::Revert => {
@@ -344,9 +345,9 @@ impl io::Write for Receipt {
                 let (result, buf) = unsafe { bytes::restore_word_unchecked(buf) };
                 let (gas_used, _) = unsafe { bytes::restore_word_unchecked(buf) };
 
-                let result = InstructionResult::from(result);
+                let result = ScriptExecutionResult::from(result);
 
-                *self = Self::script_result(result, gas_used);
+                *self = Self::script_result(result.into(), gas_used);
             }
         }
 
--- a/src/receipt/script_result.rs
+++ b/src/receipt/script_result.rs
@@ -27,7 +27,7 @@ impl From<u64> for ScriptExecutionResult {
         match value {
             0x00 => Self::Success,
             0x01 => Self::Revert,
-            0x02 => Self::Revert,
+            0x02 => Self::Panic,
             value => Self::GenericFailure(value),
         }
     }
--- a/src/receipt/script_result.rs
+++ b/src/receipt/script_result.rs
@@ -1,0 +1,34 @@
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(
+    feature = "serde-types-minimal",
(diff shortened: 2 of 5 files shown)


adlerjohn (commented):
For clarity: this only affects the field in the JSON receipt, not the canonical receipt, right?

Voxelot:
Structurally the canonical encoding is the same as before, but the difference is that the result field on a ScriptResult will have simplified values (0(success), 1 (revert), 2(panic)) instead of a complex encoding of an InstructionResult (32 bit instruction + PanicReason) crammed into a u64.

src/receipt/receipt_std.rs
```diff
@@ -169,7 +170,7 @@ impl io::Read for Receipt {
             Self::ScriptResult { result, gas_used } => {
                 let buf = bytes::store_number_unchecked(buf, ReceiptRepr::ScriptResult as Word);
 
-                let result = Word::from(*result);
+                let result: Word = (*result).into();
```
vlopes11, line 173: Is there any change here?

src/receipt/script_result.rs
```diff
@@ -0,0 +1,34 @@
...
+#[cfg_attr(
+    feature = "serde-types-minimal",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+pub enum ScriptExecutionResult {
+    Success,
+    Revert,
+    Panic,
+    // Generic failure case since any u64 is valid here
+    GenericFailure(u64),
```
vlopes11, line 11: I'm not sure its worthy to add a variant only for convert convenience. This will force the user to be aware of this generic failure when it doesn't really add a context by itself.

If the purpose of this is to increase performance, then these descriptive enums should be represented as `u8` with all variants `[0, u8::MAX]` covered. Otherwise, consider using `TryFrom` instead.

vlopes11:
I think that removing the available bits from the user is actually restricting the potential of the panic reason instead of improving it.

src/receipt/script_result.rs
Voxelot: I originally started with try from but then realized there's no rules in the specs disallowing higher values. So if someone manually serialized this value, it should still be an error right?

adlerjohn (commented):
lgtm, barring any outstanding comments by @vlopes11

Voxelot:
> I think that removing the available bits from the user is actually restricting the potential of the panic reason instead of improving it.

The panic reason is the same as before and hasn't lost any bits, on the panic receipt. It's just more transparent and enforced by the type system now.

For the ScriptResult receipt, originally we only needed to represent either success or the info about a panic. After the recent spec change noted in the original description, we now have this predicament of needing to represent _either_ a panic or a revert in the script result. There aren't enough bits in a word (`result`) to be able to entirely represent both variants, so rather than preferring the panic-centric representation (InstructionResult) over result (rA), I decided to just simplify the result. 

I agree it's less convenient when debugging a panic since you have to lookup the panic receipt instead of just looking at the script result. However the data is still there, and without changing the canonical encoding of script result (more bits / variants) there wasn't really a way to fully encode both panics and reverts into the final result. Instead of making it easy to extract panic info, but harder to extract revert `rA`, this simplification makes the control flow of scanning for the applicable receipt nearly identical in both cases for SDK users.

src/receipt/receipt_std.rs
Voxelot: nope, this is some cruft leftover from iterating on how to setup the ScriptExecutionResult.

vlopes11:
@Voxelot I refer to the encoded bits of the reason.

The original motive to use a complicated scheme to encode bits into the reason is that the VM is word-aligned - for that, the "reason" receipt will always be a `u64`. In practice, as you correctly pointed in this PR, we have only 3 possible variants (succes/revert/panic). A single byte should be enough to represent these variants, but we end up having sparing bits that will always be zeroes because in order to store this byte in the VM memory, we end up storing actually an entire word.

Since these zeroes are just available by design, the idea was to encode there additional information that can help the user to identify the cause for his panic (or revert). However, if we restrict the reason to the variants introduced in the PR, the sparing bits will never be used.

I agree with you that this "magic bits manipulation" is a complete black box, and we need a more transparent solution. One alternative is to introduce more complexity into the `Panic` and `Revert` enum variants, and create encodings to `Word`. Something like:

```
ScriptExecutionResult {
  Panic {
    cause: Something descriptive,
    ...
  }
}
```

This way, we can just consume these sparing zeroes with some useful information. We can proceed as initial improvement to set these enums, but we should in the future insert some additional complexity in this type with anything that will fit `56` bits, reserving the first `8` bits to represent the variant of the result.

src/receipt/script_result.rs
vlopes11: Hmm but then we are leaking this to the consumers. But as discussed in slack, we can proceed this way and in the future add more details to this script result type.

src/receipt/script_result.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl From<ScriptExecutionResult> for u64 {
```
vlopes11, line 14: Maybe use the type aliases created instead of plain `u64`? (`Word`)

vlopes11 (approved):
LGTM. Consider opening a good first issue to increase the granularity of `ScriptExecutionResult` into a more descriptive type

src/receipt/script_result.rs
Voxelot: done

vlopes11 (approved):
LGTM

adlerjohn (approved):

PR #105, opened 2022-04-21
Speedup Unit Tests

Voxelot:
We had some tests that took a long time to run, this saves some redundant computations taking our tests from > 1m to ~17s.

### DIFF
--- a/tests/valid_cases/transaction.rs
+++ b/tests/valid_cases/transaction.rs
@@ -201,22 +201,21 @@ fn max_iow() {
         .gas_limit(MAX_GAS_PER_TX)
         .maturity(maturity);
 
-    let secrets: Vec<SecretKey> = (0..MAX_INPUTS as usize - builder.inputs().len())
-        .map(|_| SecretKey::random(rng))
-        .collect();
+    let secret = SecretKey::random(rng);
 
     let asset_id: AssetId = rng.gen();
-    secrets.iter().for_each(|k| {
+
+    for _ in 0..MAX_INPUTS as usize - builder.inputs().len() {
         builder.add_unsigned_coin_input(
             rng.gen(),
-            k,
+            &secret,
             rng.gen(),
             asset_id,
             maturity,
             generate_bytes(rng),
             generate_bytes(rng),
         );
-    });
+    }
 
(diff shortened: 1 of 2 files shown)


adlerjohn (commented):
lgtm, but will defer to @vlopes11

tests/valid_cases/transaction.rs
```diff
@@ -201,22 +201,21 @@ fn max_iow() {
         .gas_limit(MAX_GAS_PER_TX)
         .maturity(maturity);
 
-    let secrets: Vec<SecretKey> = (0..MAX_INPUTS as usize - builder.inputs().len())
-        .map(|_| SecretKey::random(rng))
-        .collect();
+    let secret = SecretKey::random(rng);
```
vlopes11, line 204: Here we are reducing multiple secret keys to a single key. Is that desirable? If its a huge bottleneck to generate these keys, we can alternative generate a small amount of keys - but then if we are testing a single one, we might be preventing the tests from catching cases in which the internal keys management of the builder doesn't work

vlopes11 (changes requested):
Looks good overall; two questions

tests/valid_cases/transaction.rs
Voxelot: It's a valid concern, but I'd argue that if we're worried about testing with multiple private keys we should make a separate test for that specifically. This test is just about maxing out inputs and outputs which isn't necessarily related.

Voxelot: Looking into this more, it appears we have other cases which utilize multiple private keys through the builder. The TransactionFactory randomly chooses the number of inputs and keys when it builds transactions, and we have tests such as `iow_offset` that build 100's of random txs.

vlopes11: One of the functionalities of the transaction builder is to hold multiple keys and pair them with their correspondent input, creating also a correspondent witness signature.

This is already implemented in the previous test; I don't think its a good tradeoff to reduce the tested space to gain some ci time. If we can achieve ci gains via optimization, that is always desirable tho

Voxelot:
these optimizations in the test harness code aren't really needed after we have configurable params to set more reasonable numbers for MAX_INPUTS and MAX_OUTPUTS. #130

PR #107, opened 2022-04-23
Make offsets public

### LINKED ISSUE #98: Make offsets public
`TRANSACTION_SCRIPT_FIXED_SIZE` and `TRANSACTION_CREATE_FIXED_SIZE` should be public so they can be used in downstream tests

https://github.com/FuelLabs/fuel-tx/blob/8363a6e603d0b4726fa8ff2d9bfb4e5898895c5e/src/transaction.rs#L34-L58

ControlCplusControlV:
Closes #98 

Was reading through fuel-tx as I try to get better understanding of all of fuels packages and crates and figured this may still be worth patching?

I moved them to `consts.rs` to sit with the other public constants, and then adjusted the rest of the package as necessary, but can move them back if needed.

I did also test that they exported correctly locally

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -1,3 +1,6 @@
+use fuel_types::bytes::WORD_SIZE;
+use fuel_types::{Bytes32, Salt};
+
 /// Maximum contract size, in bytes.
 pub const CONTRACT_MAX_SIZE: u64 = 16 * 1024 * 1024;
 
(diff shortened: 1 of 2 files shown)


adlerjohn (approved):
lgtm, but would like @Voxelot to confirm.

Voxelot:
adding @vlopes11 to review - not sure if he intended these to be exposed. I do see the utility of exposing these because it allows you to do offset math without needing to use the offset apis on fully instantiated tx.

PR #111, opened 2022-04-29
Split `Input::coin` into predicate and signed

### LINKED ISSUE #110: Split `Input::coin` into predicate and signed
The input coin defined in the specifications is an aggregated type between two logical variants: signed and predicate.

When the input is expected to behave as coin, the signature validation should seek its witness counterpart and check the signature over its field - except for predicate and its data.

When the input is expected to behave as predicate, the validation will check if the owner is the hash of the predicate. The witness id is entirely ignored.

It is canonical to Rust to split these two different logical structures into different enum variants - even if we are to maintain the fuel specs canonical serialization.

vlopes11:
The input coin defined in the specifications is an aggregated type
between two logical variants: signed and predicate.

When the input is expected to behave as coin, the signature validation
should seek its witness counterpart and check the signature with its
fields as arguments = except for predicate and its data.

When the input is expected to behave as predicate, the validation will
check if the owner is the hash of the predicate. The witness id is
entirely ignored.

It is canonical to Rust to split these two different logical structures
into different enum variants - even if we are to maintain the fuel specs
canonical serialization.

Closes #110

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -152,7 +152,7 @@ impl Transaction {
 
     pub fn input_asset_ids(&self) -> impl Iterator<Item = &AssetId> {
         self.inputs().iter().filter_map(|input| match input {
-            Input::Coin { asset_id, .. } => Some(asset_id),
+            Input::CoinSigned { asset_id, .. } => Some(asset_id),
             _ => None,
         })
     }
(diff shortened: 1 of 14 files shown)


src/transaction.rs
```diff
@@ -152,7 +152,7 @@ impl Transaction {
 
     pub fn input_asset_ids(&self) -> impl Iterator<Item = &AssetId> {
         self.inputs().iter().filter_map(|input| match input {
-            Input::Coin { asset_id, .. } => Some(asset_id),
+            Input::CoinSigned { asset_id, .. } => Some(asset_id),
```
Voxelot, line 155: CoinPredicate also has an asset_id

vlopes11: https://github.com/FuelLabs/fuel-tx/pull/111/commits/5a3a7eaa970e5486a5b772b0894409a861d0ce38

adlerjohn (commented):
Looks fine to me, will defer to @Voxelot

Voxelot (approved):

Voxelot:
looks good to me also

PR #113, opened 2022-04-30
only check for duplicate utxo_id if input isn't a contract

Voxelot:


adlerjohn (changes requested):
Wait no, duplicate contacts on the inputs is invalid. The Sway repo tests were incorrectly including the same contract twice as inputs, which the node correctly rejected.

Voxelot:
Updated this. The issue is that two different contract inputs could provide a default utxoid, since it's assigned at block production time. If the transaction is validated before block malleability kicks in, then the transaction would be considered invalid even though the user-provided UtxoId's for contracts are irrelevant.

adlerjohn (approved):

PR #115, opened 2022-05-02
Revert "Split `Input::coin` into predicate and signed"

Voxelot:
Reverts FuelLabs/fuel-tx#111

This is major breaking change and we'd like to keep it on a feature branch for now so that we don't block the critical path of releasing the dev net.


vlopes11 (approved):
LGTM

adlerjohn (approved):

PR #116, opened 2022-05-03
Non-Exhaustive Validation Errors

Voxelot:
Make validation errors non-exhaustive so that additional future errors are non-breaking. Typically users only care about whether a validation error occurred or not, and don't attempt to match on every possible variant.


vlopes11 (approved):
LGTM

rakita (approved):
nice

PR #117, opened 2022-05-03
Split `Input::coin` into predicate and signed (#111)

### LINKED ISSUE #110: Split `Input::coin` into predicate and signed

vlopes11:
The input coin defined in the specifications is an aggregated type
between two logical variants: signed and predicate.

When the input is expected to behave as coin, the signature validation
should seek its witness counterpart and check the signature with its
fields as arguments = except for predicate and its data.

When the input is expected to behave as predicate, the validation will
check if the owner is the hash of the predicate. The witness id is
entirely ignored.

It is canonical to Rust to split these two different logical structures
into different enum variants - even if we are to maintain the fuel specs
canonical serialization.

Closes #110

### DIFF
--- a/src/transaction/offset.rs
+++ b/src/transaction/offset.rs
@@ -31,7 +31,7 @@ impl Transaction {
 
     /// For a transaction of type `Create`, return the offset of the data
     /// relative to the serialized transaction for a given index of inputs,
-    /// if this input is of type `Coin`.
+    /// if this input is of type `CoinPredicate`.
     pub fn input_coin_predicate_offset(&self, index: usize) -> Option<usize> {
         self.metadata()
             .map(|m| m.input_coin_predicate_offset(index))
@@ -41,7 +41,7 @@ impl Transaction {
     pub(crate) fn _input_coin_predicate_offset(&self, index: usize) -> Option<usize> {
         self.input_offset(index)
             .map(|ofs| ofs + Input::coin_predicate_offset())
-            .filter(|_| self.inputs()[index].is_coin())
+            .filter(|_| self.inputs()[index].is_coin_predicate())
     }
 
     /// Return the serialized bytes offset of the input with the provided index
--- a/src/transaction/types/input.rs
+++ b/src/transaction/types/input.rs
@@ -58,12 +58,20 @@ impl TryFrom<Word> for InputRepr {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Input {
-    Coin {
+    CoinSigned {
         utxo_id: UtxoId,
         owner: Address,
         amount: Word,
         asset_id: AssetId,
         witness_index: u8,
+        maturity: Word,
+    },
+
+    CoinPredicate {
+        utxo_id: UtxoId,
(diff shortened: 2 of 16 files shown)


vlopes11:
> Converting to draft just so we don't merge by accident!

Its safe to merge this PR tho. Its targeting a feature branch, and not master

src/transaction/types/input.rs
```diff
@@ -155,16 +181,24 @@ impl Input {
...
 
+    pub const fn contract_id(&self) -> Option<&ContractId> {
+        match self {
+            Self::Contract { contract_id, .. } => Some(contract_id),
+            _ => None,
+        }
+    }
+
     /// Return a tuple containing the predicate and its data if the input is of
     /// type `Coin`
```
leviathanbeak, line 198: if the input is of type `CoinPredicate` ?

src/transaction/types/input.rs
```diff
@@ -175,7 +209,15 @@ impl Input {
     }
 
     pub const fn is_coin(&self) -> bool {
-        matches!(self, Input::Coin { .. })
+        matches!(self, Input::CoinSigned { .. } | Input::CoinPredicate { .. })
```
leviathanbeak, line 212: would it be cleaner to have
```
self.is_coin_signed() || self.is_coin_predicate()
```
?

src/transaction/offset.rs
```diff
@@ -41,7 +41,7 @@ impl Transaction {
     pub(crate) fn _input_coin_predicate_offset(&self, index: usize) -> Option<usize> {
         self.input_offset(index)
             .map(|ofs| ofs + Input::coin_predicate_offset())
-            .filter(|_| self.inputs()[index].is_coin())
+            .filter(|_| self.inputs()[index].is_coin_predicate())
```
leviathanbeak, line 44: nit: the comment `...if this input is of type `CoinPredicate` ?

src/transaction/types/input.rs
vlopes11: https://github.com/FuelLabs/fuel-tx/pull/117/commits/b883896727de91e1013de8d6243f158ec30f7892

vlopes11: Done! Lost the track of the git commit since it was rebased

src/transaction/types/input.rs
vlopes11: Done! Lost the track of the git commit since it was rebased

src/transaction/offset.rs
vlopes11: Done! Lost the track of the git commit since it was rebased

Voxelot:
Depends on whether we need to make any more releases to support dev net launch. Although I think we could work all these changes in within the week, and the impact on the SDK shouldn't be too high since they use fuel-gql-client coin types.

Voxelot (approved):
:shipit:

PR #122, opened 2022-05-13
Update serde features to new schema

vlopes11:


kayagokalp (approved):
lgtm

Cargo.toml
```diff
@@ -31,14 +32,13 @@ alloc = ["fuel-types/alloc", "itertools/use_alloc", "serde/alloc"]
 builder = ["alloc"]
 internals = []
 random = ["fuel-crypto/random", "fuel-types/random", "rand"]
-std = ["alloc", "fuel-asm/std", "fuel-crypto/std", "fuel-types/std", "itertools/default", "rand/default"]
-serde-types = ["fuel-asm/serde-types", "fuel-crypto/serde", "fuel-types/serde-types", "serde-types-minimal", "serde/default", "std"]
-serde-types-minimal = ["fuel-asm/serde-types-minimal", "fuel-crypto/serde-types-minimal", "fuel-types/serde-types-minimal", "serde"]
+std = ["alloc", "fuel-asm/std", "fuel-crypto/std", "fuel-types/std", "itertools/default", "rand/default", "serde?/default"]
```
Voxelot, line 35: Should `"itertools/default", "rand/default", "serde?/default"` be under the default feature instead of std?

Voxelot (approved):

Cargo.toml
vlopes11: We achieve the same effect this way. We could either require `std` or `default` in this line, but `default` seems more permissive since these dependencies might add some other features themselves.

PR #123, opened 2022-05-13
Add bytecode length to tx

vlopes11:
A signed transaction must contain a bytecode length so it won't be
malleable.

Prior to this commit, it checked dynamically; but then the witnesses are
cleared before the tx id computation, so the length is lost.

This commit introduces the attribute that will concretely store the
bytecode len so it will be properly signed and checked.

### DIFF
--- a/tests/valid_cases/transaction.rs
+++ b/tests/valid_cases/transaction.rs
@@ -4,6 +4,8 @@ use fuel_tx::*;
 use fuel_tx_test_helpers::generate_bytes;
 use rand::rngs::StdRng;
 use rand::{Rng, RngCore, SeedableRng};
+
+use std::cmp;
 use std::io::Write;
 
 #[test]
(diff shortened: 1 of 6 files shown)


tests/valid_cases/transaction.rs
```diff
@@ -1015,3 +1015,69 @@ fn create() {
...
+    );
+
+    let id_a = tx_a.id();
+    let id_b = tx_b.id();
+    let id_c = tx_c.id();
+
+    // bytecode with different length should produce different id
+    assert_ne!(id_a, id_b);
+
+    // bytecode with same length and different content should produce same id
```
Voxelot, line 1081:
```suggestion
    // bytecode with same length and different content should produce same id
    // note: this isn't possible when the contract created output is set in a fully valid tx,
    //          but works here since we're only validating id behavior.
```

Voxelot (approved):

PR #126, opened 2022-05-23
Add `Contract` code

vlopes11:
Predicates owner is calculated with the same approach of a contract.

This is part of the validity check of a transaction so this code belongs
to `fuel-tx`.

### DIFF
--- a/src/contract.rs
+++ b/src/contract.rs
@@ -1,0 +1,144 @@
+use crate::{Transaction, ValidationError};
+
+#[cfg(feature = "std")]
+use crate::StorageSlot;
+
+use fuel_crypto::Hasher;
+use fuel_types::{Bytes32, ContractId, Salt};
+
+#[cfg(feature = "std")]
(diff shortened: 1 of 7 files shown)


src/contract.rs
```diff
@@ -0,0 +1,144 @@
...
+
+            bytes
+        });
+
+        Self::ephemeral_merkle_root(root)
+    }
+
+    #[cfg(feature = "std")]
+    /// Calculate the root of the initial storage slots for this contract
+    /// TODO: Use a sparse merkle tree once the implementation is available
```
adlerjohn, line 63: Can you file an issue for this?

adlerjohn (approved):

src/contract.rs
Voxelot: Is the comment correct in saying it should be SMT? Or is BMT fine?

adlerjohn: Storage root should be an SMT

PR #128, opened 2022-05-24
Update contract initial state root to SMT

vlopes11:
A sparse merkle tree is used to calculate the initial state root of a
contract.

It is done so due to the mutable nature of the state.

### DIFF
--- a/src/transaction/types/input.rs
+++ b/src/transaction/types/input.rs
@@ -240,13 +240,24 @@ impl Input {
     }
 
     #[cfg(feature = "std")]
-    pub fn is_predicate_owner_valid(owner: &Address, predicate: &[u8]) -> bool {
+    pub fn predicate_owner<P>(predicate: P) -> Address
+    where
+        P: AsRef<[u8]>,
+    {
         use crate::Contract;
 
         // TODO use as no-std as soon as a no-std merkle backend is available
         let root = Contract::root_from_code(predicate);
 
-        owner.as_ref() == root.as_ref()
+        (*root).into()
+    }
+
+    #[cfg(feature = "std")]
+    pub fn is_predicate_owner_valid<P>(owner: &Address, predicate: P) -> bool
+    where
+        P: AsRef<[u8]>,
+    {
+        owner == &Self::predicate_owner(predicate)
     }
 }
 
--- a/src/transaction/types/input.rs
+++ b/src/transaction/types/input.rs
@@ -230,6 +230,16 @@ impl Input {
             _ => None,
         }
     }
+
+    #[cfg(feature = "std")]
+    pub fn is_predicate_owner_valid(owner: &Address, predicate: &[u8]) -> bool {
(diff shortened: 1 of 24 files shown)


Voxelot:
I think we should have snapshot tests of the contract root and ID using some randomly generated sets of bytecode in order to ensure tweaks to this in the future won't majorly break things in the future.

vlopes11:
> I think we should have [insta](https://insta.rs/) snapshot tests of the contract root and ID calculation using some randomly generated sets of bytecode in order to ensure future tweaks won't cause major backward compatibility problems.

Since we just call the implementation of `fuel-merkle`, shouldn't we perform this check there? Otherwise, can we add this as pending issue so we unblock the predicates?

Voxelot:
1. I'm not sure how this is relevant to predicates, as it's related to contract ids
2. That's a big assumption to make by only testing fuel-merkle, and also isn't correct since the process of chunking and padding bytes is only done in this method and not in fuel core.

src/transaction/types/input.rs
```diff
@@ -230,6 +230,16 @@ impl Input {
             _ => None,
         }
     }
+
+    #[cfg(feature = "std")]
+    pub fn is_predicate_owner_valid(owner: &Address, predicate: &[u8]) -> bool {
+        use crate::Contract;
+
+        // TODO use as no-std as soon as a no-std merkle backend is available
+        let root = Contract::root_from_code(predicate);
```
bvrooman, line 239: Nit: 
I'd personally suggest something like:
```rust
let contract = Contract::from(predicate);
let root = contract.root();
...
```
where the `from` is provided by [this line](https://github.com/FuelLabs/fuel-tx/pull/128/files#diff-1da256e88678c7e356c71cc74bdc1bc0411aa5f9c04115869079cbe5c5ef6a98R116). This means you only have to add `root` to the public interface, and you can keep `root_from_code` part of the private implementation. Less public entry points means less functions to test.

bvrooman (commented):
I think adding tests for `root` would be a good idea here. 

The roots of sparse Merkle trees are tested in a number of scenarios inside `fuel-merkle`. I would assert that the reason to test `Contract::root` is that chunking the byte code and Merkle-izing these chunks is its own behaviour. 

Generally speaking, I am also a fan of TDD, and I like having tests for all public functions. This is also why I advocate for keeping the public interface as small as possible.

src/transaction/types/input.rs
vlopes11: The `root_from_code` is available to allow the user to define roots without reallocating.

If we use a `Contract::from` every time, then we always end up making an extra heap allocation: https://github.com/FuelLabs/fuel-tx/blob/32945270813e75567c55ef417149745c9b66c158/src/contract.rs#L23

adlerjohn (commented):
CI failing

Voxelot (approved):

PR #130, opened 2022-05-25
Refactor most constants as configurable parameters

### LINKED ISSUE #114: Configurable Validation Parameters
We need the ability to configure parameters at runtime used for validating transactions. This would involve things like:

- max byte size
- min gas fee
- min required coin inputs
- MAX_SCRIPT_LENGTH
- MAX_SCRIPT_DATA_LENGTH

Voxelot:
What?

This PR converts constants used for validating transactions into configurable parameters.

Why?

Making the parameters used for transaction validation compile time constants was overly cumbersome. We're still in the early stages and fine-tuning many of these parameters, so we need to be able to set these validation parameters at runtime rather than requiring new releases of fuel-tx (and docker image of fuel-core) every time a setting is tweaked. Moreover, each fuel-network instantiation may have different requirements.

fixes #114

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -25,6 +25,8 @@ mod id;
 #[cfg(feature = "std")]
 mod txio;
 
+pub mod consensus_parameters;
+
 pub use metadata::Metadata;
 pub use repr::TransactionRepr;
 pub use types::{Input, Output, StorageSlot, UtxoId, Witness};
@@ -76,17 +78,7 @@ impl Default for Transaction {
         // The Return op is mandatory for the execution of any context
         let script = Opcode::RET(0x10).to_bytes().to_vec();
 
-        Transaction::script(
-            0,
-            MAX_GAS_PER_TX,
-            0,
-            0,
-            script,
-            vec![],
-            vec![],
-            vec![],
-            vec![],
-        )
+        Transaction::script(0, 1, 0, 0, script, vec![], vec![], vec![], vec![])
     }
 }
 
--- a/src/transaction/consensus_parameters.rs
+++ b/src/transaction/consensus_parameters.rs
@@ -6,11 +6,11 @@ pub struct ConsensusParameters {
     /// Maximum contract size, in bytes.
     pub contract_max_size: u64,
     /// Maximum number of inputs.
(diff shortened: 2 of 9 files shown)


src/transaction.rs
```diff
@@ -76,17 +78,7 @@ impl Default for Transaction {
...
-            MAX_GAS_PER_TX,
-            0,
-            0,
-            script,
-            vec![],
-            vec![],
-            vec![],
-            vec![],
-        )
+        Transaction::script(0, 1, 0, 0, script, vec![], vec![], vec![], vec![])
```
leviathanbeak, line 81: just to double check, `gas_limit` was previously `100_000_000` now it's `1`, that's fine since it's just a default value?

src/transaction/consensus_parameters.rs
```diff
@@ -0,0 +1,63 @@
+/// Consensus configurable parameters used for verifying transactions
+
+#[derive(Copy, Clone, Debug, PartialEq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct ConsensusParameters {
+    /// Maximum contract size, in bytes.
+    pub contract_max_size: u64,
+    /// Maximum number of inputs.
+    pub max_inputs: u8,
```
adlerjohn, line 9: Since we're doing breaking changes anyways, can you go ahead and widen these to `u64` in line with https://github.com/FuelLabs/fuel-specs/issues/298?

adlerjohn:
It's a bit ugly to have to specify params each time your verify a transaction's well-formedness, but I guess the node wouldn't have to do that too much so it should be fine.

src/transaction.rs
Voxelot: This is a dummy default tx with 0 gas price and byte price, so we don't really need any gas limit here.

src/transaction/consensus_parameters.rs
Voxelot: done

Voxelot:
yeah there's only 2-3 places we run these validity checks, and you just need to pass this config struct around which we could bake into the chainspec

PR #136, opened 2022-06-06
Add `Transaction::receipts_root_offset`

vlopes11:
The receipts root will be updated at runtime in the vm. Hence, its
offset must be known for script transactions.

This function is checking for the correct variant of
`Transaction::Script` so it won't be misused for transactions that
doesn't contain receipts root offsets.

### DIFF
--- a/src/transaction/offset.rs
+++ b/src/transaction/offset.rs
@@ -2,7 +2,7 @@ use super::{TRANSACTION_CREATE_FIXED_SIZE, TRANSACTION_SCRIPT_FIXED_SIZE};
 use crate::{Input, Metadata, StorageSlot, Transaction};
 
 use fuel_types::bytes::{self, SizedBytes};
-use fuel_types::ContractId;
+use fuel_types::{Bytes32, ContractId};
 
 impl Transaction {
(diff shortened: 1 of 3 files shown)


src/transaction/offset.rs
```diff
@@ -195,4 +195,13 @@ impl Transaction {
...
     }
+
+    /// For a serialized transaction of type `Script`, return the bytes offset
+    /// of the receipts root
+    pub const fn receipts_root_offset(&self) -> Option<usize> {
+        match self {
+            Self::Script { .. } => Some(TRANSACTION_SCRIPT_FIXED_SIZE - Bytes32::LEN),
+            Self::Create { .. } => None,
+        }
+    }
```
rakita, line 206: Shouldn't this return only `usize` that is zero for `Create` and value for `Script`?

vlopes11: `0` might be misleading since it is a valid offset, and we shouldn't have any valid receipts root offset for `Transaction::Create`. As in the PR description/commit message, this design choice was taken so the consumers of this API won't use this offset erroneously.

PR #141, opened 2022-06-10
Add gas price factor consensus parameter

vlopes11:
The gas price factor will be used to convert between gas units and the
base asset value of the transaction.

This will empower the users to execute cheap/micro transactions without
being punished with high gas overheads.

Related issue: https://github.com/FuelLabs/fuel-specs/issues/342

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -1,4 +1,5 @@
 use crate::consts::*;
+use crate::ConsensusParameters;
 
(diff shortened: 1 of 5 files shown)


Voxelot (approved):
This looks like a better way of managing defaults 👍🏻

adlerjohn (approved):
utACK

PR #146, opened 2022-06-21
make ConsensusParams use defaults for missing values

Voxelot:
This will allow us to add new params in the future without requiring a chainspec update to deployed environments.


adlerjohn (commented):
utACK

adlerjohn (approved):

PR #147, opened 2022-06-22
Remove static_contracts

### LINKED ISSUE #138: Remove static contracts
https://github.com/FuelLabs/fuel-specs/pull/332

Dentosal:
Closes #138 

Follow-up: https://github.com/FuelLabs/fuel-vm/pull/155


tests/bytes.rs
```diff
@@ -691,48 +674,52 @@ fn create_input_coin_data_offset() {
...
+                        .expect("Failed to fetch offset");
+
+                    let (offset_p, _) = tx_p
+                        .input_coin_predicate_offset(last_input)
+                        .expect("Failed to fetch offset from tx with precomputed metadata!");
+
+                    assert_eq!(offset, offset_p);
+                    assert_eq!(
+                        predicate.as_slice(),
+                        &buffer[offset..offset + len][..predicate.len()]
```
Dentosal, line 721: Unsure if this is correct

Voxelot: hmm it seems like we shouldn't need `[..predicate.len()]` since `offset..offset + len` should already be enough to slice the predicate bytes.

Dentosal: For some reason `tx.input_coin_predicate_offset(last_input)` returns `8` as the length, but `predicate.len()` returns `2`. I've not yet managed to find what causes this disparency.

Dentosal: Ok, so apparently the length of a coin predicate is padded to a word-size in length calculation, but predicate bytes themselves aren't. https://github.com/FuelLabs/fuel-tx/blob/8e9dde715701f4535c07c2538c5f8ba7a4754c0a/src/transaction/types/input.rs#L226

adlerjohn: Does `predicate.len()` return the length in instructions, or byte?

Dentosal: Predicate is just a `Vec<u8>`, so the length is in bytes

vlopes11: We have an open issue to discuss that. I think it should be fixed in a dedicated PR for that issue

https://github.com/FuelLabs/fuel-specs/issues/357

vlopes11 (approved):
LGTM

PR #149, opened 2022-06-24
Add Message Input/Output & receipt

vlopes11:
Include message input and output with their respective validation rules.

This commit will introduce a stateful property to the transaction
validation. A valid transaction might not be validated anymore in
post-execution since its outputs might be mutated at runtime in the VM.

Related issue: https://github.com/FuelLabs/fuel-specs/pull/318

### DIFF
--- a/src/transaction/validation.rs
+++ b/src/transaction/validation.rs
@@ -266,6 +266,12 @@ impl Transaction {
             return Err(ValidationError::DuplicateInputContractId { contract_id });
         }
 
+        // Check for duplicated input message id
+        let duplicated_message_id = self.inputs().iter().filter_map(Input::message_id);
+        if let Some(message_id) = internals::next_duplicate(duplicated_message_id).copied() {
+            return Err(ValidationError::DuplicateMessageInputId { message_id });
+        }
+
         // Validate the inputs without checking signature
         self.inputs()
             .iter()
--- a/tests/valid_cases/input.rs
+++ b/tests/valid_cases/input.rs
@@ -382,17 +382,7 @@ fn transaction_with_duplicate_message_inputs_is_invalid() {
     let rng = &mut StdRng::seed_from_u64(8586);
     let message_id = rng.gen();
 
-    let a = Input::message_signed(
-        message_id,
-        rng.gen(),
-        rng.gen(),
-        rng.gen(),
-        0,
-        rng.gen(),
-        0,
-        generate_bytes(rng),
-    );
-    let b = Input::message_signed(
+    let message_input = Input::message_signed(
         message_id,
         rng.gen(),
         rng.gen(),
@@ -404,8 +394,9 @@ fn transaction_with_duplicate_message_inputs_is_invalid() {
     );
 
     let err = TransactionBuilder::script(vec![], vec![])
-        .add_input(a)
-        .add_input(b)
+        .add_input(message_input.clone())
+        // duplicate input
+        .add_input(message_input)
         .add_witness(rng.gen())
         .finalize()
         .validate_without_signature(0, &Default::default())


vlopes11:
Blocked by https://github.com/FuelLabs/fuel-asm/pull/78

src/transaction/validation.rs
```diff
@@ -134,19 +150,35 @@ impl Input {
 }
 
 impl Output {
+    /// Validate the output of the transaction.
+    ///
+    /// This function is stateful - meaning it might validate a transaction during VM
+    /// initialization, but this transaction will no longer be valid in post-execution because the
+    /// VM might mutate the message outputs, producing invalid transactions.
```
adlerjohn, line 157: > producing invalid transactions.

If the VM is implemented correctly (we can assume it is), then a valid transaction that's input into the VM will output a valid transaction. Anything else would be a DoS vector.

src/transaction/validation.rs
```diff
@@ -134,19 +150,35 @@ impl Input {
...
                 _ => Err(ValidationError::OutputContractInputIndex { index }),
             },
 
+            Self::Message { recipient, .. } if recipient != &Address::zeroed() => {
+                Err(ValidationError::OutputMessageRecipientNotZero { index })
+            }
+
+            Self::Message { amount, .. } if amount != &0 => {
+                Err(ValidationError::OutputMessageAmountNotZero { index })
+            }
```
adlerjohn, line 171: Why are zero-valued address and amount forbidden for message outputs? They're like variable outputs, they can be zero and if they're zero then they're not included in the UTXO set.

src/transaction/validation.rs
vlopes11: We do override the recipient and amount of the output:

https://github.com/FuelLabs/fuel-specs/blob/e0d5d8aea3f6c344ca47776403accc58b67eeb01/specs/vm/instruction_set.md?plain=1#L1417-L1420

It sounds odd that we execute a transaction (with zeroed recipient/amount) different than the one that was signed (with original recipient/amount). If a transaction is submitted with these fields filled, shouldn't it be invalid?

src/transaction/validation.rs
vlopes11: We can't execute `SMO` if the recipient is non-zero. Isn't this an invalid tx? One that cannot ever have its output consumed by its respective opcode?

https://github.com/FuelLabs/fuel-specs/blob/e0d5d8aea3f6c344ca47776403accc58b67eeb01/specs/vm/instruction_set.md?plain=1#L1401

pixelcircuits: I agree this validation check is necessary, but I'm confused why there isn't a similar check for OutputType.Variable to verify that amount is zero.

src/transaction/validation.rs
adlerjohn: It should be invalid, yes, but why is this stateful is my confusion.

vlopes11: This is ofc a matter of perspective, but that would be the "pre-execution" and "post-execution" states of a transaction. I will alter the behavior to produce valid transactions for both conditions and just force these values to zero during VM initialization (as it is literally described in the specs).

I just feel a bit uncomfortable of mutating the tx during VM initialization since it creates a magic box effect

vlopes11: https://github.com/FuelLabs/fuel-tx/pull/149/commits/411a630c2ecacba80711c038d65e03afa1506262

src/receipt.rs
```diff
@@ -136,7 +149,20 @@ impl Receipt {
...
+        data: Vec<u8>,
+        pc: Word,
+        is: Word,
+    ) -> Self {
+        let len = bytes::padded_len(&data) as Word;
+
+        Self::return_data_with_len(id, ptr, len, digest, data, pc, is)
+    }
+
+    pub const fn return_data_with_len(
```
Voxelot, line 165: Should we be publicly exposing this function that doesn't guarantee the length is padded correctly?

src/transaction.rs
```diff
@@ -408,6 +449,25 @@ impl Transaction {
...
+            Self::Script { outputs, .. } => outputs.as_mut_slice(),
+            Self::Create { outputs, .. } => outputs.as_mut_slice(),
+        }
+    }
+
+    /// Prepare the transaction for VM initialization
+    ///
+    /// # Steps
+    ///
+    /// 1) Zero all recipient and amount from message outputs
```
Voxelot, line 464: Should this also zero out variable outputs?

src/transaction/consensus_parameters.rs
```diff
@@ -420,4 +478,5 @@ pub mod default_parameters {
     pub const MAX_PREDICATE_DATA_LENGTH: u64 =
         ConsensusParameters::DEFAULT.max_predicate_data_length;
     pub const GAS_PRICE_FACTOR: u64 = ConsensusParameters::DEFAULT.gas_price_factor;
+    pub const MAX_MESSAGE_DATA_LENGTH: u64 = ConsensusParameters::DEFAULT.max_message_data_length;
```
Voxelot, line 481: Since these are deprecated, we probably don't want to add new constants here?

src/transaction/id.rs
```diff
@@ -84,6 +49,11 @@ impl Transaction {
 
             Output::Change { amount, .. } => *amount = 0,
 
+            Output::Message { recipient, amount } => {
```
Voxelot, line 52: is this duplicating `prepare_init`?

src/transaction/internals.rs
```diff
@@ -48,6 +49,24 @@ impl Transaction {
     pub fn set_bytecode(&mut self, bytecode: Witness) -> Option<()> {
         self._set_bytecode(bytecode)
     }
+
+    pub fn inputs_mut(&mut self) -> &mut [Input] {
```
Voxelot, line 53: any reason why these should be gated as internal? Seems useful if we wanted to have the ability to edit these fields without big match blocks around the tx in the VM or client for malleation.

src/transaction/validation.rs
```diff
@@ -226,7 +253,7 @@ impl Transaction {
         let duplicated_utxo_id = self
             .inputs()
             .iter()
-            .filter_map(|i| i.is_coin().then(|| i.utxo_id()));
+            .filter_map(|i| i.is_coin().then(|| i.utxo_id()).flatten());
```
Voxelot, line 257: Do we need to check for duplicate message inputs separately here since messages don't have a UTXO ID?

src/transaction/consensus_parameters.rs
adlerjohn: Why is message data deprecated?

tests/prepared_init.rs
```diff
@@ -0,0 +1,33 @@
+use fuel_tx::*;
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+
+#[test]
+fn output_prepare_init_zeroes_recipient_and_amount() {
```
Voxelot, line 6: Since other outputs or parts of the tx may be zeroed in the future, should we clarify this test is only for message outputs?

```suggestion
fn output_prepare_init_zeroes_message_recipient_and_amount() {
```

tests/valid_cases/input.rs
```diff
@@ -198,12 +209,154 @@ fn contract() {
...
+        generate_bytes(rng),
+    );
+
+    tx.add_input(input);
+
+    let block_height = rng.gen();
+    let err = tx
+        .validate(block_height, &Default::default())
+        .err()
+        .expect("Expected failure");
```
Voxelot, line 261: Needs docs clarifying what the expected failure actually is here

```suggestion
        .expect("Expected failure");
```

Voxelot (changes requested):
Only blocking feedback from me is clarification on whether we are properly testing for duplicate message inputs (i.e. double spend).

src/receipt.rs
vlopes11: This one will allow `const` directives and is used to restore raw serialization. It means the output will always be the representation of what is serialized, without making assumptions.

src/transaction.rs
vlopes11: https://github.com/FuelLabs/fuel-tx/pull/149/commits/38ebb1a7d60015f6d2c6cda669eb71eb2a53783d

src/transaction/consensus_parameters.rs
vlopes11: Its important we always keep consistency. If we are to remove these constants, we should remove all of them in one release. The user will still receive the deprecated warning tho.

@adlerjohn we are using the `ConsensusParameter` struct to fetch these values. These constants will be removed soon.

src/transaction/id.rs
vlopes11: Its a different context. Even tho we do the same, here its related to the signature, not to the VM initialization.

src/transaction/internals.rs
vlopes11: We should use the methods and API instead, that will (or at least should) guarantee tx consistency.

Exposing the internals is not only allowing the users to use the API to create invalid transactions, but will make their code more vulnerable to breaking changes. Ideally, somewhere in the future, we should never have to `match` these enums and use only the available API.

However, if for some reason the user still wants to manipulate the shielded attributes of a transaction, he can just, knowingly, activate this feature

src/transaction/validation.rs
vlopes11: The output of `Input::utxo_id` is now `Option<&UtxoId>` instead of `&UtxoId`. This means the check will be executed only for inputs that contains utxo id

src/transaction/id.rs
Voxelot: Makes sense, clearing for signatures isn't the same as for init (eg witnesses etc)

Voxelot (changes requested):
To be specific, I think we should have a `ValidationError::TransactionInputMessageIdDuplicated` error and prevent the same message ID from being included twice on the same transaction. The other inputs types are covered by checks on utxoid & contract id.

edit: I think we don't need checks against duplicate output message ids, since those are derived instead of stored / serialized.

adlerjohn:
Preventing the same message from being spent twice on the input side is required, otherwise it can be double-spent. Output messages as @Voxelot said derive their ID from the tx ID + output index, so by construction are unique.

vlopes11:
It doesn't look like this validation is trivial. Taking the asset id as example, we don't cross-check duplicated asset ids, as in:

https://github.com/FuelLabs/fuel-vm/blob/94810b05f3a2644e9b9a05a6bef604de84bb1250/src/interpreter/initialization.rs#L93-L101

We can generate different message id if we use the same sender but different recipient - so this suggests we can't reduce the validation to duplicated message id. I suggest we open a dedicated issue to discuss that, update the specs and after update the code here.

Voxelot:
> we don't cross-check duplicated asset ids

We are trying to mitigate the risk of users including the same UTXO / spendable input multiple times, thus effectively permitting double-spending. Duplicated asset ids across coin inputs are fine, as long as they are distinct coins. I'm having difficulty understanding the problem in the init code you referenced, as this is a reduce-operation into a hashmap (guaranteed uniqueness) which accumulates value (+=) across matching asset ids.

My understanding is that the message id is already committed to by L1/bridging, and isn't malleable. Afaik the exact specifics of our bridging haven't been made public yet. However, there is internal documentation already available from the bridging team. 

To shorten the cycle here, I'm just going to push the necessary changes to quickly unblock this.

My followup question however, is if we should also have validation that the message id claimed on the input actually matches the other fields. Since the client will likely just validate the message id exists in the state set, but won't check its validity.

adlerjohn:
The bridge contract on the other chain is guaranteed to construct unique message IDs for each message, and the client is guaranteed to only insert valid messages into its message set. Checking that the message ID matches the fields is redundant due to that.

adlerjohn (approved):

Voxelot (approved):

pixelcircuits:
Just FYI, the Ethereum contract doesn't calculate the input messageId. As a gas saving measure, it emits the data required to compute the messageId, but leaves the Fuel client to calculate the actual messageId.

PR #156, opened 2022-07-01
Feat: Use in memory Merkle trees

bvrooman:
- Upgrades `fuel-merkle` to 0.3.0
- Replaces regular Merkle trees and storage maps with in-memory Merkle trees
- Uses infallible Merkle tree methods and removes `expect` calls
- Enables `Contract` methods into the no-std context

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ description = "FuelVM transaction."
 [dependencies]
 fuel-asm = { version = "0.6", default-features = false }
 fuel-crypto = { version = "0.5", default-features = false }
-fuel-merkle = { version = "0.2", default-features = false, optional = true }
+fuel-merkle = { version = "0.3", default-features = false, optional = true }
 fuel-types = { version = "0.5", default-features = false }
 itertools = { version = "0.10", default-features = false }
 rand = { version = "0.8", default-features = false, features = ["std_rng"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ rstest = "0.13"
 
 [features]
 default = ["fuel-asm/default", "fuel-crypto/default", "fuel-merkle/default", "fuel-types/default", "std"]
-alloc = ["fuel-types/alloc", "itertools/use_alloc", "serde/alloc"]
+alloc = ["fuel-types/alloc", "itertools/use_alloc", "serde?/alloc"]
 builder = ["alloc", "internals"]
 internals = []
 random = ["fuel-crypto/random", "fuel-types/random", "rand"]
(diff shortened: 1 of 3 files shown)


Voxelot (dismissed):

Cargo.toml
```diff
@@ -31,12 +31,12 @@ rand = { version = "0.8", default-features = false, features = ["std_rng"] }
 rstest = "0.13"
 
 [features]
-default = ["fuel-asm/default", "fuel-crypto/default", "fuel-merkle?/default", "fuel-types/default", "std"]
-alloc = ["fuel-merkle", "fuel-types/alloc", "itertools/use_alloc", "serde/alloc"]
+default = ["fuel-asm/default", "fuel-crypto/default", "fuel-merkle/default", "fuel-types/default", "std"]
+alloc = ["fuel-types/alloc", "itertools/use_alloc", "serde/alloc"]
```
Voxelot, line 35: Just noticed we aren't properly leaving serde disabled when only alloc is enabled

```suggestion
alloc = ["fuel-types/alloc", "itertools/use_alloc", "serde?/alloc"]
```

Cargo.toml
```diff
@@ -31,12 +31,12 @@ rand = { version = "0.8", default-features = false, features = ["std_rng"] }
...
 builder = ["alloc", "internals"]
 internals = []
 random = ["fuel-crypto/random", "fuel-types/random", "rand"]
-std = ["alloc", "fuel-asm/std", "fuel-crypto/std", "fuel-merkle?/std", "fuel-types/std", "itertools/default", "rand/default", "serde?/default"]
+std = ["alloc", "fuel-asm/std", "fuel-crypto/std", "fuel-merkle/std", "fuel-types/std", "itertools/default", "rand/default", "serde?/default"]
```
Voxelot, line 39: rand should be guarded by the `random` flag, although it's being auto-included whenever std is enabled.

```suggestion
std = ["alloc", "fuel-asm/std", "fuel-crypto/std", "fuel-merkle/std", "fuel-types/std", "itertools/default", "rand?/default", "serde?/default"]
```

bvrooman:
Thanks for the review and suggestions on the cargo config @Voxelot. Since these things were getting past CI checks, does this mean there's a gap in the CI feature flag matrix?

bvrooman:
Do we have a general release cadence for our internal repos like this? I could also create the release of fuel-tx at this time.

Voxelot:
There's no fixed cadence, however, see Slack for the immediate release plan: https://fuellabs.slack.com/archives/C01L1HUJSL8/p1656683038666739

There's a possibility this could be released as a non-breaking patch, since the merkle v0.3 types aren't exposed via any public interfaces. However, changing the method signature from `(mut storage_slots: I)` to `(storage_slots: I)` may cause some CI warning failures downstream about mut not being necessary.

Voxelot (approved):

Issue #139, opened 2022-06-06
Remove byte price


adlerjohn:
https://github.com/FuelLabs/fuel-specs/pull/333

PR #157, opened 2022-07-01
remove byte price

Voxelot:
fixes: #139 

Remove byte price field as we now have the `GTF` opcode for future proof txs and don't need to pre-emptively include fields we aren't ready for yet.


adlerjohn (approved):

vlopes11 (approved):
LGTM

PR #160, opened 2022-07-06
Add `TransactionFee`

vlopes11:
The transaction base fee is calculated from its metered len, gas limit,
gas size and consensus parameters.

This computation is performed during VM initialization to deduct its
maximum fee from the initial base asset.

The new struct is introduced to provide downstream compatibility with
expected fee computations.


Voxelot:
This is most likely superseded by checked tx max_fee

Voxelot (approved):

PR #162, opened 2022-07-07
Add precomputed metadata to checked tx

vlopes11:


Voxelot (dismissed):

adlerjohn (approved):
utACK

src/builder.rs
```diff
@@ -146,6 +148,30 @@ impl<'a> TransactionBuilder<'a> {
...
+            .expect("failed to check tx")
+    }
+
+    #[cfg(feature = "std")]
+    pub fn finalize_checked_without_signature(
+        &mut self,
+        height: Word,
+        params: &ConsensusParameters,
+    ) -> CheckedTransaction {
+        self.finalize()
```
Voxelot, line 173: as an optimization, would it make sense to skip signature generation here? Sometimes it's useful to be able to build lots random txs quickly, and we don't care about the signatures, but the sigs take up lots of time.

src/checked_transaction.rs
```diff
@@ -92,6 +104,58 @@ impl CheckedTransaction {
...
+
+    /// Update change and variable outputs.
+    ///
+    /// `revert` will signal if the execution was reverted. It will refund the unused gas cost to
+    /// the base asset and reset output changes to their initial balances.
+    ///
+    /// `unused_gas` will be added to the output that contains the base asset
+    ///
+    /// `balances` will contain the current state of the free balances
+    pub fn update_outputs<I>(&mut self, revert: bool, unused_gas: Word, balances: &I)
```
Voxelot, line 119: I'm thinking we should treat unused_gas as gas instead of a refund amount. That way we can consolidate all the logic involving gas_price_factor into fuel-tx, and fuel-vm just needs to report how much GGAS is left to this method.

src/checked_transaction.rs
```diff
@@ -92,6 +104,58 @@ impl CheckedTransaction {
...
+            ._outputs_mut()
+            .iter_mut()
+            .for_each(|o| match o {
+                // If revert, set base asset to initial balance and refund unused gas
+                //
+                // Note: the initial balance deducts the gas limit from base asset
+                Output::Change {
+                    asset_id, amount, ..
+                } if revert && asset_id == &BASE_ASSET => {
+                    *amount = self.initial_free_balances[&BASE_ASSET] + unused_gas
```
Voxelot, line 133: The following is psuedo-code, and is intended to demonstrate the thinking from here: https://github.com/FuelLabs/fuel-tx/pull/162/files#r920644324

```suggestion
                    refund = (self.gas_limit() - unused_gas).checked_mul(gas_price).div_floor(gas_price_factor);
                    *amount = self.initial_free_balances[&BASE_ASSET] + refund
```

src/builder.rs
vlopes11: That makes sense

https://github.com/FuelLabs/fuel-tx/pull/162/commits/887f5f5f3b83c5e0cda9785233810697b138a525

src/checked_transaction.rs
vlopes11: That's a good suggestion!

Updated in https://github.com/FuelLabs/fuel-tx/pull/162/commits/8c5b71f0fc098e80ea5e0479301299785e37c04f

src/checked_transaction.rs
vlopes11: Updated in the previous comment

src/checked_transaction.rs
```diff
@@ -92,6 +104,77 @@ impl CheckedTransaction {
...
+        I: for<'a> Index<&'a AssetId, Output = Word>,
+    {
+        let factor = params.gas_price_factor as f64;
+        let gas_refund = self
+            .transaction()
+            .gas_price()
+            .checked_mul(remaining_gas)
+            .ok_or(ValidationError::ArithmeticOverflow)? as f64;
+
+        let gas_refund = (gas_refund / factor).floor() as Word;
```
Voxelot, line 136: We should **never** use f64 when dealing with accounting units like gas or asset values. 

Here's how I think we should safely do this, which is similar to how to implemented the gas limit deduction for the initial free balances:

```suggestion
        let factor = params.gas_price_factor as f64;
        let unfactored_refund = (self.transaction().gas_price() as u128)
            .checked_mul(remaining_gas as u128)
            .ok_or(ValidationError::ArithmeticOverflow)?;
        let gas_refund: Word = div_floor(unfactored_refund, params.gas_price_factor as u128)
            .try_into()
            .map_err(|_| ValidationError::ArithmeticOverflow)?;
```

vlopes11: Good catch there!

I just remembered we also included the tx fee structure that should totally encapsulate this behavior so we don't have to repeat this logic somewhere else (therefore, this mistake won't happen again)

https://github.com/FuelLabs/fuel-tx/pull/162/commits/6a22b7b6c66dc9e00cd20cffeb69fb0f0e42ad10

src/transaction/fee.rs
```diff
@@ -75,6 +75,20 @@ impl TransactionFee {
             .map(|(bytes, total)| Self::new(bytes, total))
     }
 
+    /// Attempt to calculate a gas as asset value, using the price factor defined in the consensus
+    /// parameters.
+    ///
+    /// Will return `None` if overflow occurs
+    pub fn gas_to_value(params: &ConsensusParameters, gas: Word, price: Word) -> Option<Word> {
```
Voxelot, line 82: [nit] maybe call this gas_refund_value, since we only floor the rounding when calculating the refund. When doing initial balances, we do ceiling based rounding.

Voxelot (approved):
Approving since this has test coverage in the vm and we need to unblock things. But we should really have tests involving all the cases of setting output values here in fuel tx. Eg. Regular asset vs base asset (revert t/f), etc

src/checked_transaction.rs
```diff
@@ -177,12 +263,48 @@ impl AsRef<Transaction> for CheckedTransaction {
     }
 }
 
+impl AsMut<Transaction> for CheckedTransaction {
```
Voxelot, line 266: CheckedTransaction should be immutable / all state changes to the transaction should be mediated by the CheckedTransaction wrapper. Providing a mutable reference to the transaction opens a backdoor that allows the transaction to be mutated in an invalid way, thus violating the contract of CheckedTransaction. 

Instead, we should either:

1. move any mutations to safe methods managed by checked transaction
2. convert CheckedTransaction into a tx, then recheck after making the changes.

vlopes11: If was set as convenience since this is a new type and we might disrupt the dependencies downstream. Moved to feature `internals`:

https://github.com/FuelLabs/fuel-tx/pull/162/commits/c38733a2abcacef76b78386f1dc0c945c98e963e

Voxelot (approved):

PR #167, opened 2022-07-25
Add tx, input, output helper fns

vlopes11:
The newly introduced instruction `GTF` will set the internal registers
of the interpreter with metadata of the transaction structure.

In order to not leak implementation details of the transaction to the
VM, we need several new helper functions.

### DIFF
--- a/src/transaction/types/output/consts.rs
+++ b/src/transaction/types/output/consts.rs
@@ -1,0 +1,25 @@
+use fuel_types::bytes::WORD_SIZE;
+use fuel_types::{Address, AssetId, Bytes32, ContractId};
+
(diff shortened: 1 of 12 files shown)


src/transaction/types/output/consts.rs
```diff
@@ -0,0 +1,25 @@
+use fuel_types::bytes::WORD_SIZE;
+use fuel_types::{Address, AssetId, Bytes32, ContractId};
+
+pub(super) const OUTPUT_CCV_TO_OFFSET: usize = WORD_SIZE; // Identifier
```
Voxelot, line 4: What does CCV stand for?

vlopes11: Coin, change and variable. Instead of writing all three and making a super big name, and considering its only used inside the module, I reduced it a bit

PR #170, opened 2022-07-28
Add `TxPointer` to input coin and contract

vlopes11:
related: https://github.com/FuelLabs/fuel-specs/pull/275

### DIFF
--- a/src/transaction/id.rs
+++ b/src/transaction/id.rs
@@ -1,74 +1,76 @@
-use crate::{Input, Metadata, Output, Transaction, UtxoId};
+use crate::{Input, Metadata, Output, Transaction};
 
 use fuel_crypto::Hasher;
 use fuel_types::bytes::SerializableVec;
 use fuel_types::Bytes32;
 
+use core::mem;
+
 impl Transaction {
     pub(crate) fn prepare_sign(&mut self) -> &mut Self {
         self.set_receipts_root(Default::default());
 
-        let (inputs, outputs, witnesses) = match self {
-            Self::Create {
-                inputs,
-                outputs,
-                witnesses,
-                ..
-            } => (inputs, outputs, witnesses),
-            Self::Script {
-                inputs,
-                outputs,
-                witnesses,
-                ..
-            } => (inputs, outputs, witnesses),
-        };
+        self._inputs_mut().iter_mut().for_each(|input| match input {
+            Input::CoinSigned { tx_pointer, .. } | Input::CoinPredicate { tx_pointer, .. } => {
+                mem::take(tx_pointer);
+            }
 
-        inputs.iter_mut().for_each(|input| {
-            if let Input::Contract {
+            Input::Contract {
                 utxo_id,
                 balance_root,
                 state_root,
-                ..
-            } = input
-            {
-                *utxo_id = UtxoId::default();
-                balance_root.iter_mut().for_each(|b| *b = 0);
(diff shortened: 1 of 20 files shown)


src/transaction/id.rs
```diff
@@ -1,74 +1,76 @@
...
                 state_root,
+                tx_pointer,
                 ..
-            } = input
-            {
-                *utxo_id = UtxoId::default();
-                balance_root.iter_mut().for_each(|b| *b = 0);
-                state_root.iter_mut().for_each(|b| *b = 0);
+            } => {
+                mem::take(utxo_id);
```
Dentosal, line 25: nit: when not using the return value of `mem::take`, I find using `*value = Default::default();` to be more readable

src/transaction/types/tx_pointer.rs
```diff
@@ -0,0 +1,155 @@
...
+        if buf.len() < Self::LEN {
+            return Err(bytes::eof());
+        }
+
+        // Safety: buf len is checked
+        let (block_height, buf) = unsafe { bytes::restore_word_unchecked(buf) };
+        let (tx_index, _) = unsafe { bytes::restore_word_unchecked(buf) };
+
+        self.block_height = block_height as u32;
+        self.tx_index = tx_index as u16;
```
Dentosal, line 100: Is there a possibility that these `as` casts are sinlently truncating?

tests/offset.rs
```diff
@@ -35,9 +35,9 @@ fn tx_offset() {
     let mut tested_output_contract_created_id = false;
     let mut tested_output_recipient = false;
 
-    let cases = 3;
+    let cases = 2;
 
-    TransactionFactory::from_seed(1296)
+    TransactionFactory::from_seed(1295)
```
Dentosal, line 40: Why are these changed?

vlopes11: That is a magic touch indeed, but its just that the factory with a seed `1295` will result in only two transactions covering all the specified cases

src/transaction/types/tx_pointer.rs
vlopes11: The original thought is that it might be overkill to check that since block height is `u32` everywhere in the stack, as `tx_index` is `u16`. Our canonical serialization only encode/decode words/`u64`, so we need to truncate

But you're right; better safe than sorry: https://github.com/FuelLabs/fuel-tx/pull/170/commits/1368be316e15c67e01a9fb43fff7f4324b103fc9

src/transaction/id.rs
vlopes11: isn't the shorter the better? I like `mem::take(utxo_id);` because of the reduced verbosity, and the std api is clear on what it is doing (e.g. taking the value)

Dentosal: Explict is better, `take` is intended for a specific situation. Docs are clear, yes, but with =default I wouldn't have to check the docs at all. Anyway, if you feel like this is better, we can just go with it.

tests/offset.rs
Dentosal: Maybe add a comment explaining that?

vlopes11: https://github.com/FuelLabs/fuel-tx/pull/170/commits/d62afa10d3ba2de8f6496eca42ba0674c21ccedc

Dentosal (approved):

PR #172, opened 2022-08-01
fix insta tests

Voxelot:
bind_to_scope isn't working as expected, use different system for insta configuration


vlopes11 (approved):
LGTM

PR #174, opened 2022-08-01
Change input predicate offset to message & coin

### LINKED ISSUE #166: Include message predicate metadata into the transaction
Currently, only coin predicates are tracked in the transaction metadata. However, we also need to support message predicate metadata.

We could consider consolidating the predicate metadata between coin and message input types, since the data that needs to be tracked is the same for both input types.

vlopes11:
Resolves #166

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -12,12 +12,28 @@ use alloc::vec::Vec;
 pub struct TransactionBuilder {
     tx: Transaction,
 
+    should_prepare_script: bool,
+    should_prepare_predicate: bool,
+
     // We take the key by reference so this lib won't have the responsibility to properly zeroize
     // the keys
     sign_keys: Vec<SecretKey>,
 }
 
 impl TransactionBuilder {
+    fn with_tx(tx: Transaction) -> Self {
+        let should_prepare_script = false;
+        let should_prepare_predicate = false;
+        let sign_keys = Vec::new();
+
+        Self {
+            tx,
+            should_prepare_script,
+            should_prepare_predicate,
+            sign_keys,
+        }
+    }
+
     pub fn create(bytecode: Witness, salt: Salt, storage_slots: Vec<StorageSlot>) -> Self {
         let mut tx = Transaction::create(
             Default::default(),
@@ -33,9 +49,7 @@ impl TransactionBuilder {
 
         tx._set_bytecode(bytecode);
 
-        let sign_keys = Vec::new();
-
-        Self { tx, sign_keys }
+        Self::with_tx(tx)
     }
 
     pub fn script(script: Vec<u8>, script_data: Vec<u8>) -> Self {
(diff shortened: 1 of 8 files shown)


tests/bytes.rs
```diff
@@ -858,10 +858,10 @@ fn script_input_coin_data_offset() {
                         );
 
                         let (offset, len) = tx
-                            .input_coin_predicate_offset(offset)
+                            .input_predicate_offset(offset)
```
Voxelot, line 861: we should update this test to include both coin and message predicates

Voxelot:
do we still need `Input::coin_predicate_len`? Seems like we only need `Input::predicate_len` now.

vlopes11:
> do we still need `Input::coin_predicate_len`? Seems like we only need `Input::predicate_len` now.

True, should be either deprecated or removed. Since we are already breaking the API, I just removed them:

https://github.com/FuelLabs/fuel-tx/pull/174/commits/e8faf6ca42a12dbbb984dbb41671b170b4eef0eb

tests/bytes.rs
```diff
@@ -728,11 +745,25 @@ fn create_input_coin_data_offset() {
...
+                        &buffer[offset..offset + len][..predicate.len()]
+                    );
+
+                    let (offset, len) = tx
+                        .input_predicate_offset(input_message_idx)
                         .expect("Failed to fetch offset");
 
                     let (offset_p, _) = tx_p
-                        .input_coin_predicate_offset(last_input)
+                        .input_predicate_offset(input_message_idx)
```
Voxelot, line 766: I just realized this test is named `create_input_coin_data_offset`. We should probably rename it or split the message offsets into their own tests.

tests/bytes.rs
```diff
@@ -858,10 +889,10 @@ fn script_input_coin_data_offset() {
                         );
 
                         let (offset, len) = tx
-                            .input_coin_predicate_offset(offset)
+                            .input_predicate_offset(offset)
                             .expect("Failed to fetch offset");
 
-                        assert_ne!(predicate.len(), len);
+                        assert_ne!(bytes::padded_len(&predicate), predicate.len());
```
Voxelot, line 895: Do we need to test message predicate offsets for scripts too? Looks like it was only added for the create tx offset test.

tests/bytes.rs
vlopes11: Wouldn't be strictly false to have that as generic: https://github.com/FuelLabs/fuel-tx/pull/174/commits/7ea93d5c0711493c2fa459d04b2009785f785ae6

tests/bytes.rs
vlopes11: The offset is not bound to the transaction variant. We can ofc test that as sanity check, but in practice it won't increase the robustness of the tests

src/builder.rs
```diff
@@ -12,12 +12,28 @@ use alloc::vec::Vec;
...
+    fn with_tx(tx: Transaction) -> Self {
+        let should_prepare_script = false;
+        let should_prepare_predicate = false;
+        let sign_keys = Vec::new();
+
+        Self {
+            tx,
+            should_prepare_script,
+            should_prepare_predicate,
+            sign_keys,
```
Dentosal, line 33: Nit: consider more concise initialization without let bindings

Voxelot (approved):

PR #175, opened 2022-08-02
Fix issues with tx init procedures

Voxelot:
Remove faulty logic for script init, as the values like stateRoot and balanceRoot are derived from storage/state and not the tx format itself. Furthermore, the utxoid should be unique for each contract input, instead of being set to the currently executing tx.

Updated specs to try to clarify some of this: https://github.com/FuelLabs/fuel-specs/pull/394

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -470,29 +470,15 @@ impl Transaction {
     }
 
     /// Prepare the transaction for VM initialization for script execution
-    ///
-    /// The callback argument is to expect a contract id and return its balance root and state root
     #[cfg(feature = "std")]
-    pub fn prepare_init_script<F>(&mut self, mut f: F) -> io::Result<&mut Self>
-    where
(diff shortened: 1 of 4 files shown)


src/transaction.rs
```diff
@@ -470,29 +470,15 @@ impl Transaction {
...
-        let id = self.id();
-
-        let (inputs, outputs) = match self {
-            Transaction::Script {
-                inputs, outputs, ..
-            }
-            | Transaction::Create {
-                inputs, outputs, ..
-            } => (inputs, outputs),
+    pub fn prepare_init_script(&mut self) -> io::Result<&mut Self> {
```
vlopes11, line 474: nit: Add a comment that the tx is expected to be sent to the VM with the balance/state root set by the client in the inputs

vlopes11 (approved):

PR #180, opened 2022-08-23
Add from/to str impls for Transaction

vlopes11:
This is required to parse transaction from logs of the VM so this
process can be automated.

Its naturally possible to also leak transaction internals to consumers
of this library so they can serialize//deserialize the transaction
themselves, but its preferable to have a canonical implementation that
can be consumed downstream.


src/transaction/display.rs
```diff
@@ -0,0 +1,24 @@
+use crate::Transaction;
+
+use core::fmt;
+use core::str::FromStr;
+
+impl fmt::Display for Transaction {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // its not worthy to extend the API requirement of serde because the transaction
```
mitch-fuel, line 8: what do you mean by worthy?

vlopes11: True, its not clear. The idea behind that is the fact that serde implements all serializations as fallible because you might have some attribute that can't be serialized (maybe pointers?)

That is not the case for a transaction, so we can just safely ignore that. I rephrased the comments to clarify that.

src/transaction.rs
```diff
@@ -176,6 +176,31 @@ impl Transaction {
...
+    /// Convert the type into a JSON string
+    ///
+    /// This is implemented as infallible because serde_json will fail only if the type can't
+    /// serialize one of its attributes. We don't have such case with the transaction because all
+    /// of its attributes are trivially serialized.
+    ///
+    /// If an error happens, a JSON string with the error description will be returned
+    #[cfg(all(feature = "serde", feature = "alloc"))]
+    pub fn to_json(&self) -> alloc::string::String {
+        serde_json::to_string(self).unwrap_or_else(|e| alloc::format!(r#"{{"error": "{}"}}"#, e))
```
bvrooman, line 188: Does this introduce downstream effects where the function consumer now has to check for the error json?

bvrooman (approved):
Just one question inline

Voxelot (approved):
lgtm

mitch-fuel (approved):

PR #181, opened 2022-08-31
remove owner from input message and message predicate

### LINKED ISSUE #179: Remove owner field on MessageInput types
Remove owner field from `MessageSigned` and `MessagePredicate`. Use recipient field for signature verification instead.

ref: https://github.com/FuelLabs/fuel-specs/issues/399

leviathanbeak:
closes #179


src/transaction.rs
```diff
@@ -420,7 +416,7 @@ impl Transaction {
                     ..
                 }
                 | Input::MessageSigned {
-                    owner,
+                    recipient: owner,
```
Voxelot, line 419: Since there's already a separate method to fetch the recipient, it is confusing to return a field for messages here which shouldn't exist. `.owner()` should return None for messages.

src/builder.rs
```diff
@@ -127,16 +127,17 @@ impl TransactionBuilder {
...
-        recipient: fuel_types::Address,
         nonce: Word,
         amount: Word,
         data: Vec<u8>,
     ) -> &mut Self {
         let pk = secret.public_key();
-
         self.sign_keys.push(secret);
+
+        let recipient = Input::owner(&pk);
```
vlopes11, line 137: This is a bit confusing. Should the recipient be the owner?

src/transaction/types/input.rs
```diff
@@ -254,8 +248,12 @@ impl Input {
         match self {
             Self::CoinSigned { owner, .. }
             | Self::CoinPredicate { owner, .. }
-            | Self::MessageSigned { owner, .. }
-            | Self::MessagePredicate { owner, .. } => Some(owner),
+            | Self::MessageSigned {
+                recipient: owner, ..
```
vlopes11, line 252: If we are to move one to the other, shouldn't we rename the attributes as well? Otherwise, it might be a bit confusing to handle the concepts.

Is the purpose of this to not cause a breaking change?

src/builder.rs
Voxelot: Basically yes, https://fuellabs.slack.com/archives/C01L1HUJSL8/p1661889305009839?thread_ts=1661889033.348139&cid=C01L1HUJSL8

src/transaction.rs
leviathanbeak: doing this now breaks two tests, since they return None when `sign_inputs` with `Input::MessageSigned` we end up with empty `witness` data

src/transaction.rs
```diff
@@ -443,11 +439,6 @@ impl Transaction {
                     owner,
                     witness_index,
                     ..
-                }
-                | Input::MessageSigned {
-                    owner,
```
Voxelot, line 448: we need to validate the recipient as if it was the owner

src/transaction/types/input.rs
Voxelot: This seems confusing to me as well, I think we should return none here.

Voxelot (approved):

Voxelot:
lgtm, although maybe bump fuel-asm to v0.9.0 in this PR as well?

pixelcircuits:
Seems ok to me

Voxelot (approved):

PR #186, opened 2022-09-21
In some function missed `MessageSigned` and `MessagePredicate` functionality

xgreenx:
In the tests for https://github.com/FuelLabs/fuel-core/pull/632, I found the handling of the `MessageSigned` and `MessagePredicate` inputs is missing in some methods.

### DIFF
--- a/src/checked_transaction.rs
+++ b/src/checked_transaction.rs
@@ -17,7 +17,7 @@ use core::{borrow::Borrow, ops::Index};
 use core::mem;
 use std::io::{self, Read};
 
-const BASE_ASSET: AssetId = AssetId::zeroed();
+pub const BASE_ASSET: AssetId = AssetId::zeroed();
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 // Avoid serde serialization of this type. Since checked tx would need to be re-validated on
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -1,3 +1,4 @@
+use crate::checked_transaction::BASE_ASSET;
 use crate::consts::*;
 
 use fuel_asm::Opcode;
@@ -158,6 +159,7 @@ impl Transaction {
             Input::CoinPredicate { asset_id, .. } | Input::CoinSigned { asset_id, .. } => {
                 Some(asset_id)
             }
+            Input::MessagePredicate { .. } | Input::MessageSigned { .. } => Some(&BASE_ASSET),
             _ => None,
         })
     }
@@ -222,6 +224,11 @@ impl Transaction {
                 Input::CoinPredicate {
                     owner, predicate, ..
                 } => Some((owner, predicate)),
+                Input::MessagePredicate {
+                    recipient,
+                    predicate,
+                    ..
+                } => Some((recipient, predicate)),
                 _ => None,
             })
             .fold(true, |result, (owner, predicate)| {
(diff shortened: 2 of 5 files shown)


src/transaction/types/input.rs
```diff
@@ -234,6 +235,8 @@ impl Input {
         }
     }
 
+    // TODO: Maybe we need to move `ResourceId` from `fuel-core` into `fuel-tx`(or `fuel-type`) and
```
xgreenx, line 238: If not, I will remove this TODO, if yes, I will create an issue=)

src/consts.rs
```diff
@@ -1,5 +1,7 @@
 use fuel_types::bytes::WORD_SIZE;
-use fuel_types::{Bytes32, Salt};
+use fuel_types::{AssetId, Bytes32, Salt};
+
+pub const BASE_ASSET: AssetId = AssetId::zeroed();
```
vlopes11, line 4: If this is becoming a public constant, then it should be added as associated constant of `AssetId`. ie `AssetId::BASE_ASSET` (or equivalent name)

src/transaction/types/input.rs
vlopes11: That depends on whether or not the resource id is going to be an attribute of the transaction. If yes, then it should be moved here

src/transaction/types/input/repr.rs
```diff
@@ -25,7 +25,7 @@ impl InputRepr {
     pub const fn owner_offset(&self) -> Option<usize> {
         match self {
             Self::Coin => Some(INPUT_COIN_OWNER_OFFSET),
-            Self::Message => None,
+            Self::Message => Some(INPUT_MESSAGE_RECIPIENT_OFFSET),
```
vlopes11, line 28: This isn't really an owner but a recipient. I don't recommend we mix things because we create a magic box effect (that we don't know what exactly is happening internally in the function, provided only its name)

vlopes11 (changes requested):
Couple of nits and suggestions

src/consts.rs
xgreenx: https://github.com/FuelLabs/fuel-types/pull/46

src/checked_transaction.rs
```diff
@@ -4,6 +4,7 @@
 //! This allows the VM to accept transactions that have been already verified upstream,
 //! and consolidates logic around fee calculations and free balances.
 
+use crate::consts::BASE_ASSET;
```
Voxelot, line 7: This is a good idea, although we should probably go one step further and eventually move this into the `ConsensusParameters`. This way it could be configured at runtime for different chains.

src/transaction/types/input.rs
```diff
@@ -247,9 +250,10 @@ impl Input {
     pub const fn input_owner(&self) -> Option<&Address> {
         match self {
             Self::CoinSigned { owner, .. } | Self::CoinPredicate { owner, .. } => Some(owner),
-            Self::MessageSigned { .. } | Self::MessagePredicate { .. } | Self::Contract { .. } => {
-                None
+            Self::MessageSigned { recipient, .. } | Self::MessagePredicate { recipient, .. } => {
```
Voxelot, line 253: Same thing as Victor mentioned below, we shouldn't mix recipient and owner behind the same getter.

src/transaction.rs
```diff
@@ -231,11 +237,17 @@ impl Transaction {
 
     #[cfg(feature = "std")]
     pub fn check_predicate_owner(&self, idx: usize) -> bool {
```
Voxelot, line 239: Hmm, it seems we don't have any tests for predicate owner validation here or in the VM :/

xgreenx:
@Voxelot @vlopes11 Waiting for you=)

src/transaction.rs
Voxelot: Since we are fixing behavior that wasn't properly unit tested before, I think we should include unit tests to cover the predicate ownership checks with this fix.

vlopes11 (dismissed):

Voxelot (approved):

PR #188, opened 2022-09-23
add gas info to checked transactions

Voxelot:
Also include unpriced gas info on checked txs, useful for block production to ensure txs don't exceed the block gas limit

### DIFF
--- a/src/checked_transaction.rs
+++ b/src/checked_transaction.rs
@@ -29,10 +29,8 @@ pub struct CheckedTransaction {
     initial_free_balances: BTreeMap<AssetId, Word>,
     /// The block height this tx was verified with
     block_height: Word,
-    /// Max potential fee
-    max_fee: Word,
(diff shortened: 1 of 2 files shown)


src/checked_transaction.rs
```diff
@@ -467,7 +467,7 @@ mod tests {
         // verify available balance was decreased by max fee
         assert_eq!(
             checked.initial_free_balances[&AssetId::default()],
-            input_amount - checked.max_fee
```
xgreenx, line 470: Why in some places function call, in some not?=)

xgreenx (approved):

bvrooman (approved):

src/checked_transaction.rs
Voxelot: I used pub fn's when they were available, but AvailableBalances is a private type and doesn't have any accessor fns.

PR #189, opened 2022-09-23
v0.20.1

Voxelot:
bump version to v0.20.1 to release max_gas and min_gas for CheckedTransaction.


Voxelot:
while this does _technically_ include a breaking change to the constructor for `TransactionFee`, that fn is never used outside of fuel-tx.

Dentosal (approved):
LGTM, patch-only increment is ok as per the comment above

Issue #194, opened 2022-09-29
Add contractId to panic receipt


Voxelot:
In order to enable access-list estimation of ContractIds by the SDK, the panic receipt needs to include an optional contractId field as specified here: https://github.com/FuelLabs/fuel-specs/pull/409

As noted in the specs, this field should not be included in canonical serialization. It's only an application-level concern and shouldn't be validated as part of the blockchain state. (ie. exclude from derive(eq) etc)

PR #195, opened 2022-10-03
Add contract_id to Receipt::Panic 

Salka1988:
Closes https://github.com/FuelLabs/fuel-tx/issues/194#issue-1391462723

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ internals = []
 random = ["fuel-crypto/random", "fuel-types/random", "rand"]
 std = ["alloc", "fuel-asm/std", "fuel-crypto/std", "fuel-merkle/std", "fuel-types/std", "itertools/default", "rand?/default", "serde?/default"]
 # serde is requiring alloc because its mandatory for serde_json. to avoid adding a new feature only for serde_json, we just require `alloc` here since as of the moment we don't have a use case of serde without alloc.
-serde = ["alloc", "dep:serde", "fuel-asm/serde", "fuel-crypto/serde", "fuel-types/serde", "serde_json"]
+serde = ["alloc", "dep:serde", "fuel-asm/serde", "fuel-crypto/serde", "fuel-types/serde", "serde_json", "std"]
 
 [[test]]
 name = "test-bytes"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,14 @@ repository = "https://github.com/FuelLabs/fuel-tx"
 description = "FuelVM transaction."
 
 [dependencies]
+derivative = "2.2.0"
 fuel-asm = { version = "0.9", default-features = false }
 fuel-crypto = { version = "0.6", default-features = false }
 fuel-merkle = { version = "0.4", default-features = false }
 fuel-types = { version = "0.5", default-features = false }
 itertools = { version = "0.10", default-features = false }
-rand = { version = "0.8", default-features = false, features = ["std_rng"], optional = true }
 num-integer = { version = "0.1", default-features = false }
+rand = { version = "0.8", default-features = false, features = ["std_rng"], optional = true }
 serde = { version = "1.0", default-features = false, features = ["alloc", "derive"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"], optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/FuelLabs/fuel-tx"
 description = "FuelVM transaction."
 
 [dependencies]
-derivative = "2.2.0"
+derivative = { version = "2.2.0", default-features = false }
 fuel-asm = { version = "0.9", default-features = false }
 fuel-crypto = { version = "0.6", default-features = false }
 fuel-merkle = { version = "0.4", default-features = false }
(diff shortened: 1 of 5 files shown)


mitch-fuel (approved):
lgtm

src/receipt/receipt_std.rs
```diff
@@ -72,12 +72,19 @@ impl io::Read for Receipt {
...
+                pc,
+                is,
+                contract_id,
+            } => {
                 let buf = bytes::store_number_unchecked(buf, ReceiptRepr::Panic as Word);
 
                 let buf = bytes::store_array_unchecked(buf, id);
                 let buf = bytes::store_number_unchecked(buf, *reason);
                 let buf = bytes::store_number_unchecked(buf, *pc);
+                let buf = bytes::store_array_unchecked(buf, contract_id);
```
xgreenx, line 87: The issue says the next:

> As noted in the specs, this field should not be included in canonical serialization. It's only an application-level concern and shouldn't be validated as part of the blockchain state. (ie. exclude from derive(eq) etc)

You don't need to do sterilization and deserialization. So better to make the field `Option<ContractId>` and fill it with `None`.

Salka1988: Jup, ypu are right. At the beginning, I did it like that, but I couldn't serialize the Optional, so I left it like this, hoping that someone would give me an idea :D So Thanx!!

src/receipt.rs
```diff
@@ -53,6 +53,7 @@ pub enum Receipt {
         reason: InstructionResult,
         pc: Word,
         is: Word,
+        contract_id: Option<ContractId>,
```
xgreenx, line 56: Issue says:
> As noted in the specs, this field should not be included in canonical serialization. It's only an application-level concern and shouldn't be validated as part of the blockchain state. (ie. exclude from derive(eq) etc)

`Receipt` derives `PartialEq`, `Eq`, and `Hash`. And it seems we don't need to take into account optional fields.

Also, maybe it is related to serde.

But I don't know because `Option<Metadata>` in the `Transaction` has the same problem=) 
image


@Voxelot @vlopes11 Do we need to do the manual implementation for the `Receipt` and `Transaction` of all derived traits?

leviathanbeak: we can just provide `#[serde(skip)]` to the fields that we don't want to include in (de)serialization

freesig: You could use [Derivative](https://github.com/mcarton/rust-derivative) to do `#[derivative(Eq="ignore")]` etc. It does add a dependency but it's a pretty minimal proc_macro dep and it avoids issue where manually impl `PartialEq` doesn't match the manual impl `Hash` etc.

xgreenx: @Salka1988 Could you also do the same for `Transaction.metadata` please?

Salka1988: Added!

src/transaction.rs
```diff
@@ -53,6 +55,7 @@ pub enum Transaction {
         inputs: Vec<Input>,
         outputs: Vec<Output>,
         witnesses: Vec<Witness>,
+        #[derivative(PartialEq = "ignore", Hash = "ignore")]
         metadata: Option<Metadata>,
```
xgreenx, line 59: We have the same field in the `Create` variant, could you add it there too please?<3

src/receipt.rs
```diff
@@ -182,8 +185,20 @@ impl Receipt {
...
     }
 
-    pub const fn panic(id: ContractId, reason: InstructionResult, pc: Word, is: Word) -> Self {
-        Self::Panic { id, reason, pc, is }
+    pub const fn panic(
+        id: ContractId,
+        reason: InstructionResult,
+        pc: Word,
+        is: Word,
+        contract_id: Option<ContractId>,
```
xgreenx, line 193: Do we need this field here? Maybe it is better to leave `None` by default and set it in the places where we can do that. It will avoid a lot of `None` in tests=)

Cargo.toml
```diff
@@ -11,18 +11,20 @@ repository = "https://github.com/FuelLabs/fuel-tx"
...
 itertools = { version = "0.10", default-features = false }
-rand = { version = "0.8", default-features = false, features = ["std_rng"], optional = true }
 num-integer = { version = "0.1", default-features = false }
+rand = { version = "0.8", default-features = false, features = ["std_rng"], optional = true }
 serde = { version = "1.0", default-features = false, features = ["alloc", "derive"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"], optional = true }
 
 [dev-dependencies]
 bincode = { version = "1.3", default-features = false }
+derivative = { version = "2.2.0", default-features = false }
```
xgreenx, line 27: Why do we need it in two places?

src/receipt.rs
Salka1988: Now the value is `None` by default, and we can change it using `set_panic_contract_id`

src/receipt.rs
```diff
@@ -347,7 +368,7 @@ impl Receipt {
             Self::Call { id, .. } => Some(id),
             Self::Return { id, .. } => Some(id),
             Self::ReturnData { id, .. } => Some(id),
-            Self::Panic { id, .. } => Some(id),
+            Panic { id, .. } => Some(id),
```
leviathanbeak, line 371: All variants start with `Self::*` except for `Panic` any special reason why ?

src/receipt.rs
```diff
@@ -183,7 +187,24 @@ impl Receipt {
...
+        Panic {
+            id,
+            reason,
+            pc,
+            is,
+            contract_id: None,
+        }
+    }
+
+    pub fn set_panic_contract_id(mut self, _contract_id: Option<ContractId>) -> Self {
```
leviathanbeak, line 199: Maybe a (slightly) nicer name would be `with_` instead of `set_` since you are consuming and returning it, instead of setting it.

xgreenx (approved):

leviathanbeak (approved):

Voxelot (approved):

PR #196, opened 2022-10-04
Added `Mint` transaction without tests.

xgreenx:


src/transaction.rs
```diff
@@ -298,58 +337,65 @@ impl Transaction {
...
-            Self::Script { metadata, .. } => metadata.as_ref(),
-            Self::Create { metadata, .. } => metadata.as_ref(),
+            Self::Script { metadata, .. }
+            | Self::Create { metadata, .. }
+            | Self::Mint { metadata, .. } => metadata.as_ref(),
         }
     }
 
-    pub const fn salt(&self) -> Option<&Salt> {
+    pub const fn salt(&self) -> Result<&Salt, Error> {
```
Voxelot, line 346: We shouldn't use Result for a "maybe" result, these should be `Option` like it was before.

xgreenx:
I will do this change in another way https://github.com/FuelLabs/fuel-tx/pull/197

PR #198, opened 2022-10-10
Return contract id in receipt panic

Salka1988:
Return the contract_id so it can be visible in the error message.
 ```rust 
Panic {
        id: ContractId,
        reason: InstructionResult,
        pc: Word,
        is: Word,
        #[derivative(PartialEq = "ignore", Hash = "ignore")]
        contract_id: Option<ContractId>,
    }
```
I had to extend `io::Read for Receipt` and `io::Write for Receipt` as well as
 `variant_len_without_data (..) {..}` for ` ReceiptRepr::Panic` , to be able to display the `contract_id` in `Receipt::Panic`.


Voxelot:
The std::io serialization is what we use for "canonical" spec serialization which was explicitly disallowed for this change so this PR is incorrect.

The reason why the contract id won't be available on the GraphQL API is due to the use of the "OpaqueReceipt" type in `fuel-gql-client` instead of a `TransparentReceipt`. This is because, up until the addition of this field, the canonical serialization matched the transparent version.

Since dry_run is the only API that needs this extra metadata, we should only need to change the `dry_run` api in the `fuel-gql-client` to use the transparent receipt type instead of opaque receipts.

PR #199, opened 2022-10-13
add getter function for contract_id

Salka1988:
I added :
```rust 
    pub const fn contract_id(&self) -> Option<&ContractId> {
        match self {
            Self::Panic { contract_id, .. } => contract_id.as_ref(),
            _ => None,
        }
    }
```
in order to use it in `tx/receipt.rs` to extend `schema.sdl` with the new `contract_id' field.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
+extern crate core;
 
 pub mod consts;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,6 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
-extern crate core;
 
 pub mod consts;
 
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -1,4 +1,5 @@
 use alloc::vec::Vec;
+use core::fmt::Debug;
 use derivative::Derivative;
 use fuel_asm::InstructionResult;
 use fuel_types::bytes::{self, padded_len_usize, SizedBytes, WORD_SIZE};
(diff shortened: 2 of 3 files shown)


src/receipt/receipt_std.rs
```diff
@@ -422,11 +422,9 @@ impl io::Write for Receipt {
 impl bytes::Deserializable for Receipt {
     fn from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let mut instance = Self::ret(Default::default(), 0, 0, 0);
-
```
Voxelot, line 425: [nit] these changes seem unrelated to the PR?

src/lib.rs
```diff
@@ -9,6 +9,7 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
+extern crate core;
```
Voxelot, line 12: why do we need this?

https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#an-exception
> [core](https://doc.rust-lang.org/core/index.html): Usually this is not necessary, because core is automatically imported, unless the crate is marked with [#![no_core]](https://github.com/rust-lang/rust/issues/29639). For example, some of the internal crates used by the standard library itself need this.

src/receipt.rs
```diff
@@ -1,4 +1,5 @@
 use alloc::vec::Vec;
+use core::fmt::Debug;
```
Voxelot, line 2: Also unnecessary, the proc-macro for debug is included in the libcore v1 prelude:

https://doc.rust-lang.org/core/prelude/v1/index.html
> pub use crate::fmt::macros::[Debug](https://doc.rust-lang.org/core/fmt/macro.Debug.html);

src/receipt/receipt_std.rs
Salka1988: Nah!

Voxelot (approved):

PR #202, opened 2022-10-25
Implementation of the `Mint` transaction

xgreenx:
Implemention of the `Mint` transaction as described in the https://github.com/FuelLabs/fuel-specs/pull/406 with modifications from https://github.com/FuelLabs/fuel-specs/issues/421.

## History
First implementation https://github.com/FuelLabs/fuel-tx/pull/196. After we decided to do a refactoring https://github.com/FuelLabs/fuel-tx/pull/197 to avoid `Option` everywhere.

Close https://github.com/FuelLabs/fuel-tx/issues/193


src/transaction/types/mint.rs
```diff
@@ -0,0 +1,327 @@
...
+
+#[cfg(feature = "std")]
+impl crate::UniqueIdentifier for Mint {
+    fn id(&self) -> Bytes32 {
+        if let Some(MintMetadata { id, .. }) = self.metadata {
+            return id;
+        }
+
+        let mut clone = self.clone();
+        // TODO: Wait for answer https://github.com/FuelLabs/fuel-specs/issues/421#issuecomment-1290410587
```
xgreenx, line 93: Depends on https://github.com/FuelLabs/fuel-specs/issues/421#issuecomment-1290410587

src/builder.rs
```diff
@@ -268,29 +270,142 @@ impl<Tx: Buildable> TransactionBuilder<Tx> {
...
+
+        let mut tx = core::mem::take(&mut self.tx);
+
+        tx.precompute();
+
+        tx
+    }
+}
+
+impl<Tx: field::Outputs> TransactionBuilder<Tx> {
```
Voxelot, line 285: This is a nice use of field-level traits 👍🏻

src/transaction/types/mint.rs
```diff
@@ -0,0 +1,327 @@
...
+/// The definition of the `Mint` transaction from the specification:
+/// https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_format.md#transactionmint
+///
+/// This transaction can be created by the block producer and included in the block only by it.
+#[derive(Default, Debug, Clone, Derivative)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derivative(Eq, PartialEq, Hash)]
+pub struct Mint {
+    /// The location of the transaction in the block.
+    pub(crate) tx_pointer: TxPointer,
```
Voxelot, line 77: Is this a overloading of usecases for TxPointer? Seems like it's supposed to only apply to inputs.

Voxelot: It is kind of a strange case because, the txpointer should be implied from the block producer and it's needed to generate the ID, but it's not included on the mint tx according to the specs. 

Seems like maybe we should update the tx format in the specs to include the txpointer since it's part of the id? And in that case, maybe it should also be included in the canonical serialization?

xgreenx: I planned to do this spec change in the scope of https://github.com/FuelLabs/fuel-specs/issues/421. 

Sorry, maybe I didn't describe it well=( Initially, the plan was to include two fields, `block_height` and `tx_index` as fields of `Mint` transaction to make it unique. And if we are okay with it, I will make corresponding modification in the spec=)

`TxPointer` is a representation of those two fields, so it fits well here, but it requires actualization in the specification.

adlerjohn: Interesting. I wouldn't call this overloading of the TxPointer type; it's a pointer to a transaction in a block after all!

Putting it into the transaction itself would make the transaction larger strictly unnecessarily, but would avoid needing special rules for computing mint transaction IDs vs other transaction IDs. I'm actually inclined to go with this proposal since it means one less rule for computing IDs.

Voxelot: resolved: https://github.com/FuelLabs/fuel-specs/pull/423

Voxelot (approved):

